PR #839, opened 2014-12-05
o.c.opibuilder: Don't default OPI Runtime perspetive switching to true

mfurseman:
Remember the users choice when dismissing the 'Open OPI in runtime perspective' dialogue, instead of assuming the user wishes to switch to that perspective.


berryma4:
fixes #840

PR #877, opened 2015-01-12
Change default property of fonts in *GraphModel classes.

ayemos:
So that there's less hard coded stuff and resolved issue about font name
displayed on property perspective.

### DIFF
--- a/applications/plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/model/IntensityGraphModel.java
+++ b/applications/plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/model/IntensityGraphModel.java
@@ -366,7 +366,7 @@ public class IntensityGraphModel extends AbstractPVWidgetModel {
 					MediaService.DEFAULT_FONT));
 			break;
 		case TITLE_FONT:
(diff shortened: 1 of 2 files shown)


berryma4:
Is there an open issue for this?

ayemos:
This bug-fix refers to this issue. #886

berryma4:
@shroffk

kasemir:
This is good. `MediaService.DEFAULT_BOLD_FONT` is really the macro `Default Bold`, much better than "Arial 12 Bold".

berryma4:
yes, ... but some people's screens will look different?

kasemir:
On Linux, when I create a new IntensityGraph, the X axis title font starts out as  "1|Arial|9.0|1|GTK|1".
.. but I don't have any Arial on that system, so it displays with a default font.
With this fix, DEFAULT_BOLD would be a change, but actually closer to the originally intended FontData("Arial", 9, SWT.BOLD).

I understand that people who aligned their displays to the pixel won't like it, but going forward I think we need to concentrate on compatibility based on functionality, not looks.
Labels will show the same text, text updates will show the same PV name, alarm sensitive widgets will remain alarm sensitive. The look of a Meter, the axis of a plot, the exact alarm-sensitive behaviour won't remain exactly the same.

PR #878, opened 2015-01-12
Fix bug: The length of polyline can be unexpectedly longer on OPI

ayemos:
Fix bug: The length of polyline can be unexpectedly longer on OPI


berryma4:
Is there an open issue for this?

ayemos:
Actually not, should I make new one now?

berryma4:
We usually create an issue, then a pull request referencing the issue.

ayemos:
OK, sorry to bother you.
I opened new issue. Here it is. #885

berryma4:
Thank you!  The point is, sometimes an issue needs more pull requests after another issue is found.  So, this way, We have one issue that can reference the many pull requests.

Issue #896, opened 2015-01-20
Support http:// in opibuilder action that opens files


kasemir:
*.opi files can be fetched from network links like http://, but when using an action to "Open File", pointing to a PVTable or Data browser config via http://, this fails with "This action can only open file from workspace or local file system."

PR #897, opened 2015-01-20
Bug896

kasemir:
Allow the OpenFileAction to open URLs in addition to workspace and local file system locations.

Uses the org.csstudio.utility.file plugin, so updated the opibuilder.feature as well.


kasemir:
Fixes #896

Issue #880, opened 2015-01-14
Memory Leak in BOY


jbobnar:
There is a serious memory leak in BOY. Any OPI that is opened is retained in memory forever. An easy way to see this is to open and close a high-memory-consuming opi (...opibuilder.imagelib/demo/PID_Water_Cooling.opi is a good example). If you repeat the open/close many times, you can see the memory consumption going up. Using a profiler, one can see that the OPIRunner and other opi elements are being retained in memory.

There are 2 problems:
1. In the whole chain of classes and delegates (like OPIRunner and OPIRuntimeDelegate) some of the things are never disposed of. 
2. Every contribution to the toolbar is retained in the ToolBarManagerRenderer until shutdown.

The problem 1. is fairly straightforward to fix and already fixes 99% of the leak. One just need to make sure that the dispose calls are properly forwarded.
The problem 2. is more difficult. Every time when an OPI is opened the OPIRunnerToolbarContributor regenerates the toolbar items, the items are added to the ToolBarManagerRenderer (modelToContribution) and are never removed. I think the definition of the toolbar items should be done in the plugin.xml (through extensions), similar to how it is done in the scantree case.

kasemir:
1. In the whole chain of classes and delegates (like OPIRunner and OPIRuntimeDelegate) some of the things are never disposed of. 
--
I wonder if this also happens with the code from the boy_runtime_view branch.

jbobnar:
There is no difference with boy_runtime_view branch. Drilling deeper it appears that the root cause is in eclipse. The bug is similar to this one: https://bugs.eclipse.org/bugs/show_bug.cgi?id=206584

After opening a new WorkbenchWindow the instance is referenced from all over the place and never released. No solution found yet.

jbobnar:
I found at least 3 bugs related to this issue
https://bugs.eclipse.org/bugs/show_bug.cgi?id=206584
https://bugs.eclipse.org/bugs/show_bug.cgi?id=436225
https://bugs.eclipse.org/bugs/show_bug.cgi?id=354538
The problem can be easily reproduced with any version of eclipse of 3.4 (perhaps even earlier) and later. 

It appears that we will have to wait at least until eclipse 4.5.

The commit be62294 disposes of the OPI related objects, which brings the memory leak to ~10-100 kB for every new workbench window that is opened and closed.

PR #901, opened 2015-01-22
Fix for issue #880

jbobnar:
This pull request should bring down the memory leak to as little as possible in the current eclipse version.


berryma4:
@shroffk @kasemir could you help test this.

berryma4:
addresses #880

kasemir:
I have not looked into the actual memory leak, i.e. didn't try this with jprofiler to compare the difference.
I assume Jaka did that.
I did look at the file diff and all Jaka does there looks useful: Added dispose handlers, removing old listeners before adding new ones.

Going forward I prefer introducing the Java 8 Optional<.> whenever we have null checks, because that documents what might be null and reduces chances of forgetting a null check.
For updates to existing code, especially something like the opibuilder where nobody right now has the complete understanding, that can be impractical when "let's just update this to use Optional<DisplayOpenManager>" requires changes all over the place. So I'm not saying this change should use Optional<>, just want to encourage keeping that in mind.

PR #902, opened 2015-01-22
Bug898

### LINKED ISSUE #898: vtype.pv: Support reading elements of pva structure
Recent EPICS V4 developers meeting agreed on syntax for accessing elements of a pvData structure.
Implement this in pvtype.pv for pva:/..

kasemir:
Updates vtype.pv to the current release of EPICS V4, meaning it no longer uses 

org.epics.pvdata, org.epics.pvaccess

but instead requires

org.epics.pvDataJava, org.epics.pvAccessJava

pvmanager.pva has already been updated, so this doesn't add a new dependency but instead removes a dependency on outdated V4 plugins.

Then it fixes #898, i.e. implements access to structure elements via

  pva://channel/structure/substruct/element

or

  pva://channel?request=field(structure/substruct/element)


kasemir:
Wait with merge, there's an access restriction getting the pvAccess client context..

kasemir:
OK, fixed the access restriction, no longer trying to directly access ClientContextImpl

Issue #899, opened 2015-01-21
remove utility.pv


kasemir:
alarm server, archive engine, scan server, PV table, EPICS PV Tree use vtype.pv.
opibuilder, probe use (via simplepv) either utility.pv or PVManager.

Remove utility.pv; have simplepv select between vtype.pv or PVManager.

kasemir:
On the bug899_util_pv branch,  utility.pv dependencies have been removed from the opibuilder (via simplepv).

The older o.c.diag.probe plugin has also been updated to use vtype.pv instead of util.pv, but otherwise left as is since products tend to include o.c.diag.probe.pvmanager anyway.

The only code that still depends on the utility.pv.* plugins is now `org.csstudio.archive.reader.archiverecord`. It doesn't seem worth updating, however, since it doesn't return any data. Its `ArchiveRecordReader` class is where Jan left it 2 years ago with a TODO to return VType values.

@berryma4 : Exclude archiverecord from pom.xml so it's not built, then remove all  utility.pv.* plugins?

berryma4:
That sounds fine with me.
Thank you!

Issue #818, opened 2014-11-14
Repo clean up for easing development 


berryma4:
Main two issues:
 - need to build only what changed
 - timestamp, qualifier, snapshot issue

shroffk:
module creation - Issue_818 branch in the main cs-studio repo for preparing the restructured repo
Archive (FRIB, BNL)
Databrowser (SNS)
Scan (SNS)
Shift (BNL)
OpiBuilder (SNS)
Utility (BNL)
Diag (BNL)

features
- new features if necessary
- clean up features (included plugins)

https://openepics.ci.cloudbees.com/view/CS-Studio-Applications/
create cloudbees jobs to build these modules.

kasemir:
@shroffk @berryma4 Since I'm done with my list of opibuilder, scan, databrowser, I've moved archive to my list as well because its my code after all. If you're already in the middle of doing that one, then of course proceed, otherwise I'll likely handle it early next week.

shroffk:
Please feel free to move the archiver too.
I am planning to work on my part early next week (this week I have been distracted with the connections and macro fun stuff...)

shroffk:
NOTE: git follows renamed files, the git history and blame should continue to show all previous commits. It is not possible to follow the changes at a folder level.
If the need be the history can be modified with filter-branch --tree-filter .....

berryma4:
Why are: 
org.csstudio.alarm.beast.test
org.csstudio.apputil.test
org.csstudio.apputil.ui.test
org.csstudio.apputil.ui
org.csstudio.apputil
org.csstudio.opibuilder.alarm

not in respective folders?

berryma4:
We really need to close this out please.
The applications/plugins directory needs to be empty.
Whatever is there on Monday, I will be moving to applications/unorganized.  So we can close this project and release 4.1.

shroffk:
These applications don't exactly fall in any of the modules already set up.

kasemir:
I'll move org.csstudio.alarm.beast.test while fixing #1095.

With org.csstudio.opibuilder.alarm I wasn't sure where it belongs, since it depends on both org.csstudio.opibuilder and org.csstudio.beast.alarm

berryma4:
I vote for org.csstudio.opibuilder.alarm to be with opibuilder :+1:

kasemir:
I can move org.csstudio.opibuilder.alarm into applications/opibuilder.
[...]

PR #909, opened 2015-01-29
Bug899 util pv

kasemir:
Progresses on #899, removing all dependencies to utility.pv that I could find.

The plugins org.csstudio.utility.pv, ...ui, ..pv.epics, ..pv.simu are left in place for now in case some site-specific tools still require them, but all known dependants have been updated:

 * opibuilder: Its simplepv was able to select either pvmanager or utility_pv. Now it offers pvmanager or vtype_pv, and the latter has been checked to support all that's needed on SNS beam lines. Noted some opibuilder write actions that started but never stopped PVs, fixed that.
 * probe: the older o.c.diag.probe has been updated to use vtype.pv (but the probe.pvmanager version is by now the better choice)
 * org.csstudio.archive.reader.archiverecord: Was commented in pom.xml so it's not built. Was used by DESY, but for about 2 years it didn't actually provide any data, just "return null".
 * o.c.autocomplete.ui and o.c.utility.pvmanager.ui used to check utility.pv.PVFactory.getSupportedPrefixes(), now they check vtype.pv.PVPool.getSupportedPrefixes if that is available. To allow these core plugins to access vtype.pv, it had to be moved from applications to core.


kasemir:
Wait.. after building succeeded a few times, now I have a pvAccessJava and pvDataJava version error ..

berryma4:
You want this before or after the #818 merge for core?

Issue #912, opened 2015-02-04
BOY: Connector between linking containers is not correctly loaded


ayemos:
There's a problem with loading/saving Connection Models from/to opi file.

The problem can be reproduced by trying to create Connection between a widget inside LinkingContainer and a widget outside, and to {save it and load it, go to runtime mode, and so on...}.

The issue is about the way BOY loads LinkingContainer from opi file which first tries to just create LinkingContainer itself and load linked opi file after all other widget(includes connection models) are loaded.
In this way, some connections which have one or two endpoints inside linking container can't find those widgets on endpoints when it is created so it can't be initialized and aren't shown on editor and/or runtime mode.

As an addition I made changed to fix the problem on trying to load LinkingContainer which has load-loop in .opi file hierarchy. Before, it just looked for a violation of loading itself(load A.opi from a LinkingContainer on A.opi) so it wasn't able to detect cycle: A.opi loads B.opi which loads C.opi which loads A.opi...

kasemir:
Did the merge resolve the issue?

kasemir:
Too late now, but for the record this was a bad idea.

The content of a linking container is not meant to be known outside of the linking container.
The parent *.opi cannot make any assumptions about widgets inside a linking container, because the content of the linked file can change. Only the macros required by the linked file will hopefully stay constant as the exact content of the linked *.opi is modified.
In addition, many users want to select a different file for a linking container at runtime. 

--> The widgets inside a linking container must remain private to the linking container. Connectors from the parent *.opi to widgets inside the linking container are conceptually not possible/permitted.

Unfortunately, the original implementation of the widget model merged the content of linking containers into one big model, so access to widgets inside the linking container _was_ possible. This update used the flaw in the original implementation to permit connectors to widgets inside linking containers. The exact implementation added `if (widget instance of LinkingContainer)` type checks which hint to a class hierarchy design problem.

kasemir:
We should close this as "won't fix".

The content of the linking container is private to the linking container. When a linking container displays "embedded.opi", the content of "embedded.opi" can be edited over time. Or the linking container is used to switch between "one_embedded.opi" and "another_embedded.opi". In any case, the display that contains the linking container cannot make any assumptions regarding the content of the linking container.

The current implementation of the BOY editor is somewhat cute in displaying the content and even allowing you to select widgets inside the container and edit them, but that's misleading because such changes will _not_ be saved to the "embedded.opi".

[...]

PR #913, opened 2015-02-04
Bug fix: Connector between linking containers is not correctly loaded 

ayemos:
Changes:

ConnectionModel.java
- Add src_path, tgt_path property to ConnectionModel so it can
distinguish two widgets which is included in two different
LinkingContainer but have same WID.
- Add loadedFromLinkedOpi attribute to prevent some connection inside
LinkingContainer from being saved to .opi file.

XMLUtil.java
- Change the order of procedure of converting XML element to widget so
it can correctly load connection between widgets loaded in
LinkingContainer.
- Add utility methods :stringToXML, inputStreamToXML.
- Add ability to detect loops in LinkingContainer and loaded .ooi files.

AbstractLinkingContainerModel.java,
AbstractLinkingContainerEditPart.java
- Put LinkingContainer up from org.csstudio.opibuilder.widgets to
org.csstudio.opibuilder so those classes in org.csstudio.opibuilder can
handle LinkingContainer Models/EditPart separately from ordinal
Container Model/EditPart.


ayemos:
#912

PR #932, opened 2015-02-09
added diirt features to build

berryma4:
#818


berryma4:
commented out pva because of import errors

Issue #800, opened 2014-11-05
OSGI diirt bundles (Lead: Gabriele - Reviewers: Eric, Kay)


berryma4:
@berryma4 will do the service loader part

berryma4:
For the record, this update to 'master' has been reverted. The issue_800 branch has the corresponding changes.

berryma4:
Current todo:
  - [ ] continue moving plugins from org.epics.vtype to org.diirt.vtype
  - [ ] continue moving plugins from org.epics.utils to org.diirt.utils.*
  - [ ] continue moving plugins from org.epics.pvmanager.* to org.diirt.datasource.* (and similar; source)
  - [ ] resolve service extension point issue
  - [ ] resolve configuration issue

gcarcassi:
The current to do is enable a product build on the cs-studio-4-integration branch. There is going to be no progress until that is done.

berryma4:
currently, pvmanager is taken from a clone for the pva change in the upstream diirt.
artifacts are uploaded to maven.controlsystemstudio.org.

shroffk:
These points are now part of the #1171 Issue

PR #938, opened 2015-02-10
moving to diirt 3.0

berryma4:
#800


kasemir:
For the record, this update to 'master' has been reverted. The issue_800 branch has the corresponding changes.

PR #946, opened 2015-02-12
Renamed new core features for consistency; 

berryma4:
moved startup, startup.helper, openfile to utility
Also, separated core utility feature into a rcp and rap


berryma4:
#818

Issue #967, opened 2015-02-19
RAP Databrowser and OPI no longer work


css-iter:
The webdatabrowser and webopi no longer work. Two reasons:
1. Refactorization of the databrowser plugin, which now uses a different charting library
2. General plugin reorganization. Some plugins have been removed and some new dependencies introduced.

kasemir:
Thanks for updating the RAP build!

kasemir:
I think the original issue had been resolved, but as the opibuilder and databrowser code just moved into their own applications/* subdirs, might be good to confirm that the RAP build survived that change.

PR #968, opened 2015-02-19
Making RAP work again (Issue: #967)

css-iter:
Updated the code to make the RAP applications work again. Fixes the issue #967 

A new plugin org.csstudio.rap.rtplot has been created, which is a copy of the org.csstudio.swt.rtplot and modified only as much as absolutely necessary for RAP. The solution is not ideal, because 99% of the code is duplicated. Eventually, we need to do something similar as we have with xygraph.rap anx xygraph.rcp. However, we should wait until RAP 3.0 gets released, because they promise some new features that will get used by the rtplot.

All RAP features have also been updated to match the newest plugin dependency trees.


berryma4:
thank you!

Issue #982, opened 2015-02-25
org.csstudio.autocomplete.logbook could move from applications/plugins to applications/logbook/logbook-plugins


richardfearn:
It depends on `org.csstudio.autocomplete` (which is in `core`) and `org.csstudio.logbook` (which is of course already in `applications/logbook/logbook-plugins`).

`org.csstudio.logbook.viewer` depends on it.

It's currently included in `org.csstudio.applications.all.feature`.

berryma4:
Yes!!  Please do help with the reorg project: #818
@shroffk is on vacation, but that's no reason for the fun to stop.

PR #984, opened 2015-02-26
Move org.csstudio.autocomplete.logbook from applications/plugins to applications/logbook/logbook-plugins (#982)

richardfearn:


berryma4:
related to #818

berryma4:
closes #982

PR #987, opened 2015-02-26
Cleanup whitespace in {ui,utility}-repository/pom.xml so they match base/platform/unorganized

richardfearn:
Makes the 5 POMs more similar


berryma4:
#818

PR #996, opened 2015-03-02
Bug977

### LINKED ISSUE #977: BOY changed how it resolves paths to displays opened in actions
We have displays created with 3.x that are linked as follows:

main.opi has linking container for ../some/place/other.opi.
That other.opi than has an action button for ./final.opi, located in that same ../some/place, i.e. in the same directory as other.opi.

Used to work fine, because the relative path to the action was resolved based on the location of where it's defined, i.e. other.opi.
Now it's resolved based on when it's called, i.e. relative to main.opi, and then fails because there is no final.opi in the directory of main.opi.

kasemir:
Fixes #977 

#913 performed quite some change on the linking container, introducing a new AbstractLinkingContainerModel and how the linking container appears in the widget tree.
It added several `if (.. instanceof *Linking*)` checks to account for the special handling of the linking container.
This adds one more to handle the resolution of relative links. Stuff within the linked content is again resolved relative to the location of that content.

### DIFF
--- a/applications/plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/model/AbstractWidgetModel.java
+++ b/applications/plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/model/AbstractWidgetModel.java
@@ -1,22 +1,22 @@
-/* 
- * Copyright (c) 2006 Stiftung Deutsches Elektronen-Synchroton, 
+/*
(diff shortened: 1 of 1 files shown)


berryma4:
A lot off line ending changes

kasemir:
Yes, my 'save' actions remove trailing spaces, add @Override etc.

PR #997, opened 2015-03-04
Updated all the ITER copyright headers from '2010-2014' to '2010-2015'.

jbobnar:
Entire pull request is about updating the copyright header in all files that contain the ITER copyright.


kasemir:
I've always wondered if this is necessary.
A google search once left me thinking "Copyright 2010" with just the first year is good enough.
It states when you created something, and from then on you have the copyright.
Updating that each year serves no purpose.

Still, if this floats your boat, it's certainly benign enough, so should we just merge this?

jbobnar:
It's ITER's policy. Every build script, xml, source file (not just CSS) are updated every year.

PR #1002, opened 2015-03-09
Opi validation tool

jbobnar:
Implementation of the OPI validation tool. It was written according to ITER specifications, but perhaps someone else may find it useful as well. That is why I put it into applications/plugins. 

It uses the eclipse wst.validation framework to validate an OPI file against the OPI Editor Schema according to the validation rules defined in the rules.def file (see Preferences -> Validation -> OPI Validation). Every property can be either 
- read-only: the value in the schema and the value in the OPI should always be the same
- write: the value in the OPI should always be different than in the schema
- read/write: whatever.
Quick fix is available for the read-only properties.

To use the tool include the it in the product definition or in one of the features. Be careful about its dependencies.


berryma4:
thank you!

Issue #799, opened 2014-11-05
Java 8 development branch (Lead: Kunal - Reviewers: Gabriele)


berryma4:

shroffk:
Use of Instant and Period
Integration of JavaFX

shroffk:
TODO:
-use a released verison of e(fx)clipse instead of the nightly releases
-the pom files need to be updated to include the p2 repo hosting the efx plugins
-include the required features in the products to ensure that logging.ui works with the javafxviews

shroffk:
All TODOs: 
Merged to Master

Issue #734, opened 2014-10-01
Log configuration


kasemir:
The Oct. 1 hangout mentioned
"We need a log console in the product (where you can choose plugins to listen to, and the level)"
There is a related ticket about how to collect all log data, #439

I think we have most of the pieces, except the configuration is too coarse.
* We have agreed a while back to use java.util.logging which comes free with the JRE
* org.csstudio.logging can route log messages to the Console view, to JMS, to files, …
* org.csstudio.logging.ui can configure that

… but we only have the option to turn file/console/JMS logging on/off at the INFO, WARNING, .. level.
We cannot configure by individual logger, i.e. enable messages from `org.csstudio.security` while disabling those from `org.cstudio.vtype.pv`

PR #1006, opened 2015-03-12
Dev java8

shroffk:
Logging.ui with javafx


berryma4:
#799 Adding javaFX integration using e(fx)clipse to manage visibility
#799 #734 creating first javafx view using e(fx)
#734 creating a new command to reload the loggers.

Issue #1012, opened 2015-03-16
Using E4 parts in RCP


kasemir:
More a What-I-learned-at-EclipseCon than an issue, but creating issue so that it can be referenced in pull requests.

E4 introduced the application model and dependency injection. RCP was adjusted to work on top of that, and earlier Eclipse 4.x releases concentrated on simply getting it all to somewhat function again. It was not possible to add E4 parts via model fragments to an RCP app because RCP would create the complete model, eliminating any directly added E4 parts.

Good news is that the main issues have been handled. One can now add E4 parts via model fragments, in there read preferences etc. via dependency injection. As a shortcut to a complete model fragment, "e4view" can list those parts in the RCP plugin.xml as views.

PR #1013, opened 2015-03-16
o.c.util.clock: Replace ViewPart with E4 Pojo

kasemir:
Example for #1012:
'e4view' that allows using E4 Pojo in RCP. Part is now Pojo, using DI to get parent and preference setting.
Could have used the same ID as the View, but that creates problems when
existing workspace had already used View with that ID.
Command to open the view, Preference page and online help remain unchanged, they're provided by RCP on top of E4.


berryma4:
that's pretty cool :+1:

kasemir:
.. forgot to update version in pom.xml ..

Issue #544, opened 2014-05-28
BOY, mostly: Rethink Perspectives/Compact Mode/…, use Views for runtime (Lead: Kay - Reviewers: Kunal, Eric)


kasemir:
Goal is to be able to “save” the layout of *.opi displays, so it can be “restored” and operators can switch between different layouts.
Maybe including a “frozen” mode where a “restored” layout cannot be altered.
Ideas (unclear if this is the solution):
OPI Runtime always based on “View”, and “Perspective” is used to save/restore.
Update compact mode to be more compact (hide status bar, …).

This should help with #457, #438, #101

kasemir:
Notes from hangout today with FRIB, BNL, DLS, SESAME, ITER, SNS.

SNS:
See http://ics-web.sns.ornl.gov/kasemir/CSS_OPI_Views2.ppt.
All OPIRuntimes in RCP View.
Can use RCP Perspective to switch View arrangements.
Includes _all_ Views (Probe, Properties, ..).
Excludes partstack "org.eclipse.ui.editorss".

DLS:
All OPIRuntimes in top-level SWT Shell.
Running inside CSS, with usual context menu for PVs etc.
External tools can use --launcher.openFile to open these OPIRuntimes.
Since they're top-level shells, Linux/X11 tools can be used to locate them based on their window title and adjust their position.

SESAME:
See http://www.aps.anl.gov/epics/meetings/2014-10/presentations/Wednesday/Fall%202014%20EPICS%20Collaboration%20meeting_Ibrahim%20Saleh.pdf pages 12, 13.
All OPIRuntimes in standalone Eclipse Window, compact mode.
Compared to DLS, this adds the 'tab' for the single part within the window. Could display multiple tabs within a window, but don't really want to do that.
Added mechanism to load/save "Profiles": Iterates over all Shells in the workspace, identifies those which contain OPIRuntime, saves their *.opi path and macros. On restore, restores those that were saved and closed those which happen to be open but are not included in the saved profile.
Using CSS in "Development" or "Runtime" mode. Same CSS binary, but runtime mode closes/hides the main Eclipse workspace window, so end users only see the standalone Eclipse Windows with the compact-mode OPIRuntimes.

willrogers:
Changes that are needed for OPI Shell:

* OpiShell.java
* OpenOpiShellAction.java
* Add OpenOpiShellAction in WidgetActionFactory.java
[...]

PR #1019, opened 2015-03-18
Opi shell

willrogers:
As being discussed in #544.  You may find much to discuss in here - comments very welcome.

Points to note:

* I've added Require-Bundle to opibuilder, without specific versions
* OPIShell partially implements IOPIRuntime, which is inelegant but seems to work
* another javascript function is needed to correctly close all types of OPI

### DIFF
--- a/applications/plugins/org.csstudio.opibuilder/META-INF/MANIFEST.MF
+++ b/applications/plugins/org.csstudio.opibuilder/META-INF/MANIFEST.MF
@@ -33,7 +33,9 @@ Require-Bundle: org.eclipse.ui;resolution:=optional,
  org.epics.vtype;bundle-version="2.7.1",
  org.python.jython;bundle-version="2.5.2",
  org.csstudio.swt.widgets;bundle-version="2.1.0",
- org.csstudio.swt.xygraph;bundle-version="2.0.4"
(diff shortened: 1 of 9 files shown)


kasemir:
Not sure what the procedure is; usually Eric merges the pull requests, but that's for merging into 'master'.

Since this is for the proj544_1.1_runtime_in_view branch, I'd like to just do it.
Then I'd like to sync that branch back with master, but the problem there is that recent logging.ui changes introduced JavaFX and links to a 'nightly' eclipse update, which causes some difficulties.

So for now I'd like to merge this pull request as is into  proj544_1.1_runtime_in_view.
After master has been updated to point to released eclipse dependencies, sync the branch back to master. This is expected soon, within a day.

Then we can work on a preference to configure the default: Open in Workspace, Open in New Window? Update the context menu to use those new strings, see if everything else still works.

At that point, it should be usable, and could be merged into master.

willrogers:
Yes, makes sense.   Have a play around and see if you've got any comments on how it works.

There are a few other changes that we've got:

* Changes to the EDM converter that use this functionality
* A couple of EDM-style tweaks, such as copying PVs on middle-click
* Right-click option to edit OPI

These are probably separate discussions.

Issue #1016, opened 2015-03-18
Data browser: Allow ordering of traces


kasemir:
Allow moving trace up/down in list of traces

PR #1030, opened 2015-03-19
Databrowser tickets

kasemir:
Allows ordering of traces #1016 @berryma4
and adds title #1017 @shroffk


shroffk:
While this pull request has been merged I wanted to report the following test
1. Tested the Title option both in databrowser and the embedded plots.
2. Tested the re-ordering of plots, ensured the order is reflected correctly in the value Axis, the order preserved correctly in the save .plt file

kasemir:
Thanks for checking.
The ordering code is not 100% perfect:
When an item is already at the top, you still get the context menu option to move it up further.
Well, it won't go any further, since it cant, 's that's OK, but when you then push "undo", it actually moves down one.
Ideally, "Move Up" would be disabled if you can't move up (and "..down" if you're already at the bottom).

In addition, it simply moves the items up and down.
If you have this arrangement:

Item 1 on axis 1
Item 2 on axis 2
Item 3 on axis 1

-> Axis 1 shows "Item 1, Item 3"

Moving Item 3 up results in this:
Item 1 on axis 1
Item 3 on axis 1
Item 2 on axis 2

-> Axis 1 still shows "Item 1, Item 3"

Might be better to move items "within their axis" as opposed to simply moving them in the list of items.


.. but handling those two issues would be a lot more work, and compared to some of the world's REAL ISSUES I figured it doesn't justify the effort.

PR #1035, opened 2015-03-24
Opi shell

willrogers:
New behaviour as discussed in #544.  OPI shell integration is more consistent.


kasemir:
Thank you for digging into this.
I like how you changed OPEN_DISPLAY_TARGET into OpenDisplayTarget to assert that all places that used the 'old' options has indeed been updated.

willrogers:
There's still a separate OpenOPIInView action, which behaves inconsistently.  There's also an OpiRuntime perspective, which is still required when trying to open a view in a specific location. 

What's the plan for these?

kasemir:
As you say, the perspective is required.
I'll look at the OpenOPIInView action, see if that's superfluous.

When opening a *.opi in View, there used to be the idea that it checks the perspective and prompts "Do you want to switch to the OPI Runtime perspective" because that's the one which supports opening in specific locations.

.. but that was based on simply checking the perspective name. What if you are already in "MyVacuumPerspective" which has been saved based on a customization of the "OPI Runtime perspective". Not sure if it's worth hacking around to check if the current pers. is somehow based on the OpiRuntime perspective, or not. May be best to just remove that check and in general have no automatic perspective switch at all. Changing the perspective will simple be left to the user.

PR #1036, opened 2015-03-25
Opi builder validation

jbobnar:
Updates to opibuilder.validation plugin:
Fixed a few bugs, and implemented a few new features (e.g. remove rule, additional allowed values).

A few minor and non invasive changes were made in the opibuilder core. See 649aefc, a977cbc, 57bc956 for the details of these changes.


berryma4:
@shroffk @kasemir there are a few changes to boy, other than the validation plugin.  Can I get another check on this merge.

jbobnar:
I added a few commits related to validation. 
6238d45 also reverts some changes I previously made in the in the XMLUtil. It turned out I was making a workaround for some linking container and connector issues (those changes will be in another pull request).

kasemir:
Eric, good that you pick up on boy changes.

The addition in OPIEditor.java to catch ResourceException and sync the workspace seems like a good idea, but overall I think it's better to set this preference which causes Eclipse to always refresh on access in case any file is out of date, no matter if *.opi, *.plt, *.txt, *.py, ..:

org.eclipse.core.resources/refresh.lightweight.enabled=true

jbobnar:
@kasemir Thanks for the hint. 

I already saw refresh.enabled in our customization and thought that was it and that it doesn't work 100%. I removed that piece of code.

kasemir:
"refresh.enabled", I think, is the periodic refresh that helps new files to show up in the navigator.

"refresh.lightweight.enabled" is relatively new and different. It removes the "resource out of sync" errors and instead has Eclipse simply refresh that file before opening it. That helps a lot in our case where one user may update a file, then another user opened it and got that  "resource out of sync" error. I think the error is gone with "org.eclipse.core.resources/refresh.lightweight.enabled=true", which should really be the default.

jbobnar:
Are the rest of the changes to opibuilder acceptable? Can we merge this now that the resource resync has been reverted?

berryma4:
@shroffk @kasemir I'm okay with the merge, everyone else?

kasemir:
Yes, OK.

PR #1038, opened 2015-03-25
Rap update

jbobnar:
Updated the rap.rtplot to match the changes (made by Kay) in the nonRAP version.


kasemir:
Really, thank you very much for continuing to update the RAP version of things!

PR #1042, opened 2015-03-30
o.c.scan: SequenceCommand needs empty constructor

kasemir:
.. because that's called via reflection to for example put it into the
tree editor's command palette. Had deleted it because the var-arg
constructor is otherwise good enough, but turns out zero-arg constructor
is still needed.


kasemir:
Eric, you're very fast today in merging pull requests, but I'm even faster in adding bugs ;-)

berryma4:
:+1:

Issue #599, opened 2014-07-28
BOY Symbol widgets should support SVG animation


utzeln:
To display SVG image files, Apache™ Batik SVG Java-based Toolkit is currently used by the BOY Symbol Widgets.

The latest revision of Batik, release 1.7 supports a nearly complete implementation of declarative SVG animation.

Example of svg xml code for a square rotating about its centre:
```<rect x="150" y="20" width="60" height="60" fill="blue">
  <animateTransform attributeType="xml"
                    attributeName="transform"
                    type="rotate"
                    from="0 180 50"
                    to="360 180 50"
                    "begin="3s" dur="6s"
                    repeatCount="indefinite"/>
</rect>```

ITER Use Case:
In case of a valve opening/closing, an animated SVG file that displays the valve in its open then close position alternatively at 2Hz frequency has to be supported.

arnaudf-iter:
Added ```o.c.utility.batik.SVGHandler``` to handle both static & animated SVGs + updated ```o.c.swt.widgets.symbol.SVGSymbolImage``` to use it.

Some bugs in Batik framework dependencies have been fixed by Jaka adding 2 new fragments:
- ```org.csstudio.batik.dom.fragment_1.0.0.jar```
- ```org.csstudio.batik.dom.svg.fragment_1.0.0.jar```

PR #1044, opened 2015-04-03
SVG animation

jbobnar:
Implementation of animated SVG images.


berryma4:
@shroffk @kasemir BOY changes that need another tester.
Thank you!

kasemir:
org.csstudio.swt.widgets/src/org/csstudio/swt/widgets/figures/ImageBoolButtonFigure.java removes 'synchronized' keywords. I don't know how the synchronization is supposed to work there. Maybe that's fine. If in doubt, I'd feel better with added thread-safe copies. Instead of just removing the synchronized from dispose:
```
public /* no more synchronized  */ void dispose() {
    // Can onImage change while this code is running?
    if (onImage != null && !onImage.isDisposed()) {
        onImage.dispose();
        onImage = null;
```

.. do this ..:

```
public void dispose() {
    final SymbolImage copy;
    synchronized (this)
    {
      copy = onImage;
      onImage = null;
     }
     if (copy != null && ! copy.isDisposed())
		copy.dispose();
```

arnaudf-iter:
```ImageBoolButtonFigure``` was updated to use ```org.csstudio.swt.widgets.symbol.SymbolImage``` and so, be able to display SVG, GIF & PNG the same way as Symbols Widgets. To do this, I copy/pasted a lot of methods from ```org.csstudio.opibuilder.widgets.symbol.multistate.CommonMultiSymbolFigure``` which already has these 'synchronized' keywords for no particular reason as far as I know.

We shouldn't have any issues with the methods not being synchronized. Everything should be called from the GUI thread. So any synchronization is redundant.

berryma4:
does this close #599 ?

arnaudf-iter:
Yes, done.

Issue #998, opened 2015-03-05
BOY changed how it resolves scripts from *.opi files in linked containers


kasemir:
Similar to #977:

Assume you have `/path/to/shared/display_with_script.opi` that loads `some_script.py` from that same directory. Works fine.

Now you include that in a linking container from within `/other/path/main.opi`.
==> `some_script.py` not found, because it's resolved as `/other/path/some_script.py` instead of `/path/to/shared/some_script.py`

kasemir:
Found a fix, but mostly learned that the `LinkingContainer` is problematic.

In principle, it includes a different `*.opi` file, with its own widget tree.
But loading an `*.opi` file merges all linked trees into one big widget tree, as if the `LinkingContainer` was like a `GroupingContainer` within the same tree.

This has always prevented us from simply implementing "open in editor" on a runtime, because the merged runtime model is different from the editor model.
#912 tried to fix some connector issue, added more `*Linking*` classes and a bunch of `instanceof` checks. #977 fixed one follow-up issue with relative display links. This fixes an issue with script paths, unclear if that's new or was always there.

Bottom line is that the whole `LinkingContainer` handling could be redone.

PR #1052, opened 2015-04-08
Bugfix for linking container, connection

jbobnar:
With #912 some of the existing features no longer worked (e.g. relative paths were incorrectly resolved, macros were not correctly propagated, connection did not work if it started in one linking container and ended in another one etc.). The issues have been fixed, at least the ones that are covered by ITER's test plans.

In addition, the FileUtil has been updated to resolve relative paths and to support Xinclude.

### DIFF
--- a/applications/plugins/org.csstudio.opibuilder.rap/src/org/csstudio/opibuilder/util/ResourceUtilSSHelperImpl.java
+++ b/applications/plugins/org.csstudio.opibuilder.rap/src/org/csstudio/opibuilder/util/ResourceUtilSSHelperImpl.java
@@ -36,6 +36,16 @@ public class ResourceUtilSSHelperImpl extends ResourceUtilSSHelper{
 	private static final String NOT_IMPLEMENTED = 
 			"This method has not been implemented yet for RAP";
 
+	@Override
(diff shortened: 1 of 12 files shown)


kasemir:
After #912 resp. the associated pull #913, I had to fix the relative loading of scripts #998 and displays #977. Can you confirm that this additional cleanup for 912/913 still resolves scripts and displays?

The test is like this:

test1/display.opi
test1/other.opi
test1/test.py

display.opi has an action button to open "other.opi", and also something that executes "test.py".
When simply opening display.opi, those relative links to files in the same directory work fine.


Now add
test2/main.opi

with a linking container to "../test1/display.opi".

Originally, that worked fine: test2/main.opi loads ../test1/display.opi, and from then on the links inside display.opi are resolved relative to its location, i.e. ../test1/other.opi and ../test1/test.py.

After  912/913, they were instead resolved relative to the top-level test2/main.opi, so it failed when trying to load test2/other.opi and test2/test.py. 977 fixed the relative display lookup, and 998 fixed the relative script lookup.

jbobnar:
@kasemir 
I confirm your test case still works with these changes.

PR #1066, opened 2015-04-17
Proj544 1.1 runtime in view

willrogers:
This change means that if a new display is opened and Mode is 'Replace Current', the existing OPIShell will be replaced by the new OPIShell.  As discussed in #544.


kasemir:
Looks good to me, and is for proj544_* branch, not master, so merging.
Thanks!

kasemir:
@willrogers There is still some quirk in the 'Replace' handling.

* Install BOY Examples
* Right-click on 4_Actions_1.opi in Navigator, Open With, OPI Display (Standalone Window)
* Try the "Open 1_Start_Up.opi Replace" button

 -> Thanks to your update, it replaces that in the OPIShell.
Good, you're now in 1_1_Start_Up.opi.
But the "Go Home" button of 1_1_Start_Up.opi, when executed in the OPIShell, ended in a null pointer error because the display has no runtime.
I fixed that with ab951f802cafe7a6964a7f26797d50264badd658, so the "Go Home" action no longer crashes, but the navigation is now different.

When using View:
Open  4_Actions_1.opi in workbench View, press "Open 1_Start_Up.opi Replace", then "Go Home"
--> These are all replacements within the same panel.

When using Shell:
Open  4_Actions_1.opi in Shell, press "Open 1_Start_Up.opi Replace", then "Go Home"
--> First transition replaces content of OPIShell, but "Go Home" then goes back to a View because the OPIShell has no IOPIRuntime.

willrogers:
I see. That's a plumbing error that I've fixed.

I spotted one more bug (a null pointer exception) that I'll have to fix on Monday.

PR #1068, opened 2015-04-17
Issue 818

kasemir:
Moves applications/plugins/*opibuilder* and applications/features/*opibuilder* into a new applications/opibuilder module with applications/opibuilder/opibuilder-plugins, opibuilder-features and opibuilder-repository.

Some pom.xml were adjusted because their parent is now applications/opibuilder instead of just applications.

Not moved, even though they mostly belong below applications/opibuilder:

 * org.csstudio.opibuilder.alarm: Left in applications/plugins, because it depends on both org.csstudio.opibuilder, to which it's a fragment, and org.cstudio.alarm.beast, which it makes available to KEK scripts.

 * org.csstudio.swt.xygraph: Also used by scan.ui.plot (were I could remove it, but haven't, yet) and o.c.sds.components.ui.

#818


berryma4:
#818

PR #1069, opened 2015-04-17
applications/opibuilder: Add .project files

kasemir:
For applications/channels, alarms, logbook we also have the .project files, so maybe best include them for applications/opibuilder as well.


berryma4:
#818

PR #1070, opened 2015-04-20
Proj544 1.1 runtime in view

willrogers:
As discussed in #1066 and #544.

Fixes a possible null pointer exception when launching OPIShell.

Perhaps this method of pull requests is getting a little tedious, so if it's easier for you you could give me permission to push to the main repo.


kasemir:
You're right. On branches other than master, you should be able to just commit. I think I've just added you, let me know if it still doesn't work.

Issue #803, opened 2014-11-05
Alarm/OPI interaction. PVs to display & ack’ alarms (Lead: Nadine - Reviewers: Eric)


berryma4:
I have an awful version I use for one opi screen:
https://stash.nscl.msu.edu/projects/CF/repos/org.csstudio.pvmanager.jms.beast.feature/browse

kasemir:
Notes from discussion at today's hangout..

Simpler TO-DO Items:
1) Core plugin for alarm colors (foreground, background) and icons to represent the alarm severities. Then use that in alarm plugins, opibuilder, any other place that wants to render alarms.

2) Alarm Table columns: Configure via preferences similar to message log history.
Replace with JavaFX Table? Would allow easy runtime changes to columns, may overall be a better table. But note that this change is primarily about configuring the columns once, and operators shouldn't need to (be able to?) change them at runtime.

3) Default 'double-click' action for Alarm Table.
.. mostly to open one designated related display.


Alarm PVs: PV for alarm ack, table PV for current alarms, …
This once looked like a great building block.
If the Alarm Table etc. don't work for you, create your own *.opi that talks to the alarm server.
--> Impossible?
Based on Eric's trials, no good fit for 'PV'. How to support alarm-specific context menus?


Alarm filtering:
Provide a 'Contextual' Alarm Table that uses E4 'Context' Variable for current alarm hierarchy level.
Set by Alarm Tree, or set by BOY Script, or set by new ITER navigational widget that updates linking containers, ...
Alarm Table only shows alarms below that level. Its tab name reflects the currently active alarm hierarchy level. Has option to show 'all', ignoring the alarm level context variable.

Not 100% if E4 context variable is the best mechanism, but looks like it at this time.
The Lars Vogel article http://www.vogella.com/articles/Eclipse4Services/article.html lists E4 communication mechanisms.
In E3, there was mostly the ISelectionProvider, one global selection mechanism.
If you want to pass updates on various model items between E4 components, you can use context variables.
Publishing anything into a context variable is as easy as publishing a selection, but note that you can now name the item that you put into the context. This allows you to publish different types of information under different names, effectively having access to multiple selection services.
You can either use a string or a class reference: 

public void methodThatPublishesContextVariable(final MWindow window)
{
    final IEclipseContext context = window.getContext();
    context.declareModifiable("alarm_level");
    context.set("alarm_level", "/The/Current/Alarm/Level");
}

[...]

PR #1084, opened 2015-04-26
Alarm table Customization

jbobnar:
Changes related to #803.

Added the plugin org.csstudio.ui.resources.
Added the alarm table double click handler.
Implemented severity icons.
Implemented the acknowledge tick box.
Implemented alarm table columns configurator (select columns, and define order).
Added alarm action and alarm ID columns.
Implemented grouping of alarms according to acknowledge/unacknowledge state.

Missing: filtering and possibility to have multiple alarm tables


berryma4:
@kasemir are you okay with these changes (discussed in #803)

kasemir:
I thought #803 was about
 * grouping all alarms in one table, i.e. merging the separate tables for active and ack'ed tables into one
 * some shuffle-the-columns preference settings
 * icons

This change also changes the text displayed for the alarms, sometimes using the first guidance instead of the alarm name?
In addition, it introduces some extension point that links the double-click to either acknowledging alarms, displaying the guidance, or invoking commands?! What happens on a double-click needs to be consistent and can't change that way, so I'm not sure how that's helpful.

--> I'd like to try this code before we merge it.

kasemir:
The option to group yes/no, i.e. to have one table vs. separate tables for active and ack'ed alarms, is purely a runtime configuration via the table's context menu. Same for selecting the table columns and their order. Initial table width cannot be configured.
--> Need to be preference settings. When users start new instance of CSS (on new workspace), they should see the alarm table as one or as separate active/acknowledged tables as desired by site settings, configured via preferences. Doesn't need preference GUI, but preference settings that one can put into site's plugin_customization.ini are requred.

The new 'ack' column can only acknowledge and not un-acknowledge.
--> Since it's now easier to accidentally acknowledge alarms without thinking, it needs to be as easy to un-ack to correct such errors.

Alarm icons were placed in a sharable org.csstudio.ui.resources plugins
--> Alarm colors should be there as well, moving the current preference settings and GUI from alarm plugins into org.csstudio.ui.resources.

kasemir:
Found why the 'unack' doesn't work.
When clicking on the new 'Ack' column icon for an acknowledged alarm, first the new org.csstudio.alarm.beast.ui.alarmtable.GUI#AcknowledgeEditingSupport will un-acknowledge the alarm.
Then, however, the same mouse click will trigger the also newly added GUI#mouseDown() handler, which  again acknowledges the alarm.

--> Implement one of those two mechanisms only, so they don't fight each other.

jbobnar:
To explain why I did things like they're done.

Double click is done as an extension, because 1. I didn't want to force the double click for CSS users/labs that do not want this functionality, 2. I don't think all labs want "open first display" as the default double click action. If I were an operator, I'd probably want to see the history of the PV or the message history for that alarm when I double click the item. By using the extension points every lab can configure the double click as they wish. It's not configurable by user, but it can be confiugred by the developer.
Similar situation is with the Alarm ID and Alarm Action. ITER wants to display guidance title and guidance message, but someone else might want to display something else.
If you think the above usecases do not apply to anyone, let me know and I hardcode the bahaviour.

> Initial table width cannot be configured.

Are you suggesting to ditch the ColumnLayout and use predefined settings for each column? I think that is a bad GUI design, because the size of the table is defined by the CSS window size. Forcing a fixed default width would make it look strange.

> Found why the 'unack' doesn't work.
> Implement one of those two mechanisms only, so they don't fight each other.

Sorry, that was a mistake that happened while I was juggling the commits between different branches to separate pull requests by functionality.

> Alarm colors should be there as well, moving the current preference settings and GUI from alarm plugins into org.csstudio.ui.resources.

That is the plan, I just didn't get that far yet. If you prefer I can continue develop in a separate branch and create only one pull request when everything is done? However, that will not happen before June.

utzeln:
Having the alarm GUI settings as preferences in the ini file is required to setup the default site alarm table configuration. As for the message history, the idea is to provide both - preferences and a GUI to change them at runtime even if most of the time, the users won't change them. As we can have multiple alarm tables, the GUI to change the settings should be accessible from the table itself and not from CSS -> Preferences.

Regarding the column width, it is important to be able to configure it. What was not confirmed yet in  #803 is if the weight of the column has also to be taken into account as it is in the message history - it somehow adjusts also the width according to the space available.

I like the idea of extension points :)! And ITER HMI design guide is not final, so this allows more flexibility.

kasemir:
Lets first concentrate on where we have agreement and get those fixed resp. actually implemented:

* Preference setting for all-in-one vs. separate active/ack'ed alarms
-> Needs to be added. Then still allowing user to change the running instance is fine, but the very first thing needed is a preference *.ini setting.

* Column content and order
-> Should also be configurable via *.ini file. Then still allowing user to change the running instance is fine, but the very first thing needed is a preference *.ini setting.

* Column size
Original proposal suggested "Preferences should allow to specify if a column is shown or hidden...and its width.". I see no way to configure the column width, not in preferences nor via the view menu dialog where the column names and order are configured. The only way to change the column widths is via the layout as I had it from the start.
Do you want to allow configuring the column widths (and weights) via preferences as originally stated, or has this been dropped?

* Column to acknowledge alarms
-> Needs to also allow un-acknowledge. I would implement that _only_ via the table editor, not also via the GUI#mouseDown() handler which confirms in its comments that it's a hack, and didn't work after all.



After we get through the basic column config, and the 'acknowledge' actually works, we could discuss other changes like what to do on double-click, or what else to show.

A pull request that doesn't do what was originally stated, instead has bugs and introduces new things that have not been discussed is what worries me.

jbobnar:
> Preference setting for all-in-one vs. separate active/ack'ed alarms
That is done.

> Column content and order
You can already specify the list of columns and order in the preferences.ini. 
What do you mean by specifying "column content"?

> Column to acknowledge alarms
As I said I made a mistake juggling the commits. It's been updated. The reason for the chosen approach is to be able to ack/unack multiple items at once by holding shift or ctrl. 

> A pull request that doesn't do what was originally stated, instead has bugs 
It has one bug, which has been fixed.

> and introduces new things that have not been discussed is what worries me.
Please name those, so I'll know what you are talking about.

kasemir:
Hi:

OK, I'll try to find time to check the code again.

>What do you mean by specifying "column content"?
List of columns and order, which you say is done.

What about the column width & weight, which was originally mentioned but I didn't see that implemented in the preferences. Is that also done?

> introduces new things that have not been discussed
Extension points for double-click handler and tableTextProvider that can provide an arbitrary text for any table cell have not been discussed.

jbobnar:
> What about the column width & weight, which was originally mentioned but I didn't see that implemented in the preferences. Is that also done?

No, those are not done. I am not sure how that should be done and if it would bring any added value. Presetting the width would suggest to ditch the ColumnLayout. I don't want to do that. As a user I'd I want the table to take all available space, but I don't want it to go off screen. 
I was asking about that in one of the previous posts. I also haven't received any answer from ITER yet.

kasemir:
OK, so let's remove the widths from "Preferences should allow to specify .. width.":
I never liked that in the first place, which is why I had not implemented it and instead used the table column layout. You don't like it either, so it won't be there.

kasemir:
The basic ack and now also un-ack via click in the new "Ack" column works.
But ack/unack via click on the new column doesn't work for me when multiple rows are selected.
In fact trying to select/deselect multiple items now easily accidentally acks/un-acks a single item.

Compared to the GUI#mouseDown(), using a table cell 'editor' instead of a GUI is less code, less of a hack, and what the GUI#mouseDown() was meant to accomplish doesn't work.


==>  I'll create a different pull request reduced to
1) New alarm icon and Ack column
2) Support for alarm table that combines active and acknowledged alarms
3) Preferences to select initial table setup (combined? which column headers in what order?)

Once we agree on that, we can start separate discussion on DoubleClickHandler, TableTextProvider.

kasemir:
Just noticed that sorting by PV name no longer worked when alarm time was made the default.

Closing this, will submit new pull request for just the icons, combined/split tables, preferences.

jbobnar:
> what the GUI#mouseDown() was meant to accomplish doesn't work.

It actually works fine: http://users.cosylab.com/~jbobnar/multipleack.avi

> Just noticed that sorting by PV name no longer worked when alarm time was made the default.

There are a few other sorting comparators missing: current status, value, action and ID

I noticed that in the ack case we need to check the permissions. Do you want to do that in the EditingSupport, AcknowledgeAction#run() or in AlarmClientModel? I cheer for the model, because it covers all existing and future acknowledge cases at the same place.

kasemir:
Regarding  GUI#mouseDown():
Your video actually shows the problem: Clicking on icons in the new "Ack" column still selects/deselects that row.

Assume you have used shift-click and/or control-click, or the toolbar PV name filter to select a bunch of alarms, and then:

a) You use the context menu Ack/Unack entries: Selected items are acked resp. un-acked, selection stays as is.
b) You use the Ack/Unack buttons in the tool back: Selected items are acked resp. un-acked, selection stays as is.

That's good.

c) You click on one of the new "Ack" column icons: Details depend on luck and OS. At the time the mouseDown() handler runs, the previously selected items may still be selected, or have already been un-selected. So previously selected items may get acked/unacked or not. Afterwards, just the row in which you just clicked is selected.

That's not good. You just changed the acknowledgement of items which are no longer selected, or not. The primary response to clicking on anything in a row is selecting that one row, no matter if SWT table, Excel spreadsheet, .. Acknowledging an alarm must be a conscientious decision, where it's very obvious which alarms will be affected and which ones were affected. When using the context menu or the toolbar buttons, that's provided. With the column editor implementation, it's also OK: You click in one column, that one row will be affected and only that one row is afterwards selected. The mouseDown() handler does not provide enough gain to make up for its shortcomings.


Regarding the sorting:
We used to be able to sort on PV name. #1084 broke that, which was clearly a step back. I've fixed that and added support for the new ICON column, but you're right that the existing STATUS & VALUE as well as the new ACTION and ID columns are not supported (i.e. default to sorting by alarm time). I'll add them #1093. 


Icons:
We added the alarm table icons. The alarm tree always had icons, and now that's inconsistent with the alarm table icons. Alarm tree has the added requirement to reflect if some alarm trigger PVs are disabled. Need to decide how to reflect that with the new icons #1094.

Issue #1088, opened 2015-04-29
Legend for databrowser plot


shroffk:
With the new databrowser plot, where the legend is no longer included and the information is displayed on the Axis, certain configurations for the plot do not render well.

An example is show below:
![image](https://cloud.githubusercontent.com/assets/2111304/7391660/baaaf494-ee50-11e4-899c-3168595f803b.png)

In this case in addition to maximizing the plot area, since these plots are primarily used to display time series data, it is to provide a way to maximize the horizontal span too.

Proposed solution:
To add a collapsible legend.
![image](https://cloud.githubusercontent.com/assets/2111304/7391705/fccc521e-ee50-11e4-98f0-1dcb787182f1.png)

kasemir:
Basically, makes sense.

Missing/todo:
 * Add [x] Show legend to properties, persist & reload to/from config file.
 * Instead of its own gray, use plot's background color
 * Add LegendFont (so far: Title, Label, Scale)
 * Context menu "Hide Legend", "Show Legend" (renamed from "..bar") still necessary in spite of properties. Show/hide of legend will 'dirty' the editor (unless configured to ignore 'dirty' changes)
 * Code is/remains mostly in RTPlot

Later:
 * Instead of Composite/Labels/GridLayout, make this an RTPlot PlotPart that's drawn in background thread. No immediate visible difference to users, but better in the long run: Draws in background, can create/release colors instead of having to pass colors to Label, and will allow to somehow reflect which axis contains which trace.

shroffk:
Missing/todo:
* legend show update when trace display char are changed (name, color...)

kasemir:
Re: legend show update when trace display char are changed (name, color...)
Is now the case.


Re: databrowser.opiwidget should have show/hide toolbar and legend options.
Yes, at runtime, as you had already added.


Re:  .opi visibility properties should override the .plt properties for embedded plots.
Maybe not.
The toolbar visibility property was in the opiwidget because the *.plt failed to preserve that.
Now the *.plt stores the toolbar visibility (thanks to you), but unfortunately we already have opiwidgets storing it as well, so yes, the *opi needs to override the toolbar visibility from the *.plt.
For the legend, I think we start out right: The *.plt file has the config, so the *.opi doesn't need it.

shroffk:
[...]

PR #1091, opened 2015-04-30
Issue 1088

shroffk:
@berryma4 @kasemir 
Changes were made to RTPlot to support a legend (very basic) which can be shown/hidden

Changes to the databrowser
Model updated to have a field representing the legend font, XMLPersistence saves/loads the legend info from the model
Misc tab provide UI to edit legend font


shroffk:
@kasemir could you check the use of the SWTMediaPool

berryma4:
#1088

kasemir:
I'll change the whole legend implementation from Composite/Labels/GridLayout into an RTPlot PlotPart that's drawn in background thread.
Kunal, I think I can already write to your repo, so I'll just add commits to this branch & pull request over the next couple days.

kasemir:
Finally got around to fetch the source code. I keep getting
```
java.lang.NullPointerException
	at org.csstudio.swt.rtplot.internal.LegendHandler$1.changedTraces(LegendHandler.java:52)
	at org.csstudio.swt.rtplot.internal.Plot.fireTracesChange(Plot.java:1132)
```
where the code in question is `Display.getCurrent()` being null.

--> Another reason for implementing the legend such that it's drawn in the background as a plot part, so I'll start working on that.

kasemir:
I've changed the legend implementation from Composite/Label to background-drawn within the plot.

You had added both the Toolbar and the Legend visibility to the *.plt model, so they're saved in the config file, changes trigger the 'dirty' indication of the editor. Good.

Since the toolbar visibility wasn't always saved in the *.plt file, the opiwidget had a toolbar property. In hindsight, would have been better to leave that to the *.plt file, but OK, so now the *.plt file has the toolbar property, but the opiwidget has one as well.

For the legend, I think it's best to leave that only in the *.plt file, where luckily it is from the start. So the opiwidget just shows/hides legend as per the *.plt file. Can be changed at runtime, but the opiwidget otherwise doesn't have its own show_legend property.

kasemir:
OK, let me put the opiwidget option to show/hide legend, overriding that setting from the *.plt model, back in.

shroffk:
Sorry about the NPE.

kasemir:
The opiwidget property to show/hide legend is back.
From my side, this is now good, i.e. it's what you had, with legend drawn in plot's thread.

Nice would be a legend that groups by axes, with border that matches axis color.
I don't think I'll get to that right now, and it doesn't affect the properties/config files, so we can leave that for a little later. I do think it's necessary, though, since otherwise the legend (with trace names on the axes turned off) leaves you wondering what's on which axis.

berryma4:
Are we done?

kasemir:
Yes, for now. I'll improve the legend later.

PR #1092, opened 2015-04-30
bug fix #1081 macro substitution

### LINKED ISSUE #1081: Macros substitution for linking containers broken
On the master branch changes to the linking container have disabled the correct loading of the macros from the file being linked to.
Additionally system macro are also not correctly attached to the linking container( ${LCID} does not work)

In the LinkingContainerEditpart in the configureDisaplyModel(), we see that the macroMap for the display model is correctly updated and has all the required macros, with expected values.

opi files to recreate this issue have been added to the opi-regression-tests repository.
opening the screens in 3.3.x and 4.1.x  shows the discreapancy

berryma4:
I see a lot of issues here.  I tried my best to only address the macro issue. @shroffk @kasemir could you test also.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
@@ -12,16 +12,20 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 
 import org.csstudio.opibuilder.editparts.AbstractBaseEditPart;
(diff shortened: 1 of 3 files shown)


berryma4:
#1081

kasemir:
It may be early next week that I get around to test it.
Having a build problem right now related to org.csstudio.applications.all.feature still listing the alarm plugins which are in their own submodule, and the recently added org.csstudio.ui.resources. Once I figure that out, may like to get that pulled first.

berryma4:
@shroffk the applications.all feature should be removed.  It was a catch all that shouldn't be needed with the directories split.

kasemir:
For #1095, I can try to just remove  the applications.all feature

berryma4:
Is this okay now?

kasemir:
I didn't manage to remove the applications.all feature, only shortened it.

On the macros, I haven't gotten around to test it, but if it works for Kunal and you, go ahead.

shroffk:
I have tested the opi screens I created in the opi-resgression-test repository and they are working as expected.
1. The linking container ID is being correct resolved ${LCID}
2. The Child macros are correctly resolved when there is no parent macro
3. Macros defined on the linking container override the default macros value defined in the child opi.

berryma4:
thank you!

shroffk:
These tests should not be considered exhaustive.

Issue #1094, opened 2015-05-01
Consolidate alarm icons


kasemir:
Recent update added alarm table icons. The alarm tree always had icons, and now that's inconsistent with the alarm table icons. Alarm tree has the added requirement to reflect if some alarm trigger PVs are disabled. Need to decide how to reflect that with the new icons.

utzeln:
The disabled icon state could be a dimmed version of the enabled icons - invalid, major and minor, supposing that even if the alarm is disabled, the alarm server maintains its state.
Ref. https://wiki.eclipse.org/UI_Graphics_:_Design_:_States

kasemir:
Note that the disablement can be partial.
Individual alarm trigger PVs are either fully enabled or disabled.
At other levels in the alarm tree (area, system), the icon currently indicates if one or more trigger PVs below that level are disabled. A gray background to indicate "something is disabled" while otherwise using the original icon coloring to reflect the summed state of the remaining PVs might work.

jbobnar:
Here are some ideas for the icons. Please let me know, what is acceptable.

Disabled items: 
Currently all icons have white background. We could mark the nodes that have disabled children with gray background.

Half-half icons: 

a) 
We invent icons for each and every combination of severity and latched severity. In that case we can use different combinations of the red cross, yellow triangle, checkbox. For example:
![major_ack - copy - copy](https://cloud.githubusercontent.com/assets/5602519/7565143/bf08f954-f7ed-11e4-91fc-82b449f57692.png)
Downsides:
Users would need to get familiar with lots of new icons. Not a big deal for newcomers, but could be an issue with established users. The 16px icon is also too 
small for displaying that level of detail.

b)
We show double icons, where one icon is for the latched severity and one for the current one. If they are the same or if they fall into one of the categories, for which we already have an icon (e.g. cleared but not acknowledged), we show just one icon.
![major_ack - copy 2](https://cloud.githubusercontent.com/assets/5602519/7565152/d1c2a324-f7ed-11e4-9a28-9eed3fb9edea.png)
![capture](https://cloud.githubusercontent.com/assets/5602519/7565156/dcda72be-f7ed-11e4-8ba4-783396b3cd56.PNG)
Downsides:
Maybe too colorful and more space is needed to show the tree (in case of a deep tree).

c)
We continue to use the circles and half circles in the tree. The good thing is that users are already used to this and the colors resemble the severity.

kasemir:
[...]

PR #1105, opened 2015-05-07
Alarm table

jbobnar:
Implemented the missing comparators for the alarm table.
Added disabled icons.
Updated alarm tree to use the same icons as the alarm table.

### Important
**Previously the alarm tree showed half-half icons for certain conditions. We do not have icons for all those conditions (only cleared & not acknowledged exist). For example: latched major & current minor or acknowledged major & current minor. What would you like to see for those conditions?**


berryma4:
@kasemir would you like to comment on the "half" icons?

kasemir:
We really need to discuss things before you go ahead and change stuff.

The missing comparators for the alarm table are fine, and that could be one pull request.

The icons are a different issue. I had mentioned on #1094 that the disablement tends to be partial.
Individual alarm trigger PVs are either fully enabled or disabled, but at other levels in the alarm tree (area, system), the icon currently indicates if one or more trigger PVs below that level are disabled.

The icons added in this pull request are totally grayed out, which doesn't serve the purpose. Imagine you have an "Area" or "System" in the alarm tree that has 50 PVs, 10 in Major alarm, one disabled. The icon should show that the highest alarm severity at that point in the hierarchy is "Major", with some small hint that at least one PV is disabled. Your icon mostly shows an overall gray blob, suggesting the whole subtree is disabled.

A gray _background_ to indicate "something is disabled" while otherwise using the original icon coloring to reflect the summed state of the remaining PVs might work better. But as you noted there's still a mismatch between the original simple indication of latched alarm severity vs. current PV severity, which the new icons combine in some ways but without icons for every combination.

--> The whole icon stuff first needs to be agreed upon on #1094, and then we can implement it.

PR #1106, opened 2015-05-07
Implemented missing comparators (current status, icon, value, action, id)

jbobnar:
Pull request that contains only the comparison. I'll wait with everything else until an agreement is reached!

### DIFF
--- a/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.ui.alarmtable/src/org/csstudio/alarm/beast/ui/alarmtable/AlarmComparator.java
+++ b/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.ui.alarmtable/src/org/csstudio/alarm/beast/ui/alarmtable/AlarmComparator.java
@@ -10,7 +10,9 @@ package org.csstudio.alarm.beast.ui.alarmtable;
 import java.util.Comparator;
 
 import org.csstudio.alarm.beast.AnnunciationFormatter;
+import org.csstudio.alarm.beast.SeverityLevel;
 import org.csstudio.alarm.beast.client.AlarmTreePV;
+import org.csstudio.alarm.beast.client.GDCDataStructure;
 import org.epics.util.time.Timestamp;
 
 /** Comparator (= table sorter) that compares one column of an alarm.
@@ -53,7 +55,6 @@ public class AlarmComparator implements Comparator<AlarmTreePV>
 		            return level1 - level2;
 				}
 			};
-        case ICON:
         case SEVERITY:
         	return new AlarmComparator(up)
 			{
@@ -67,6 +68,30 @@ public class AlarmComparator implements Comparator<AlarmTreePV>
 		            return level1 - level2;
 				}
 			};
+        case ICON:
+//          1. Invalid / Disconnected alarm
+//          2. Major unacknowledged alarm
+//          3. Minor unacknowledged alarm
+//          4. Alarm cleared but unacknowledged
+//          5. Major acknowledged alarm
+//          6. Minor acknowledged alarm
+            return new AlarmComparator(up)
+            {
+                @Override
+                protected int doCompare(final AlarmTreePV pv1, final AlarmTreePV pv2)
+                {
+                    final SeverityLevel sv1 = pv1.getSeverity();
+                    final SeverityLevel sv2 = pv2.getSeverity();
+
+                    final boolean pv1ClrNotAcked = pv1.getCurrentSeverity() == SeverityLevel.OK && sv1.isActive();
+                    final boolean pv2ClrNotAcked = pv2.getCurrentSeverity() == SeverityLevel.OK && sv2.isActive();
+
+                    if (pv1ClrNotAcked ^ pv2ClrNotAcked)
+                        return pv1ClrNotAcked ? (sv2.isActive() ? -1 : 1) : (sv1.isActive() ? 1 : -1);
+                    else
+                        return (sv1 == sv2) ? super.doCompare(pv1, pv2) : sv1.ordinal() - sv2.ordinal();
+                }
+            };
         case STATUS:
         	return new AlarmComparator(up)
 			{
@@ -79,6 +104,18 @@ public class AlarmComparator implements Comparator<AlarmTreePV>
 	                return super.doCompare(pv1, pv2);
 				}
 			};
+        case CURRENT_STATUS:
+            return new AlarmComparator(up)
+            {
+                @Override
+                protected int doCompare(final AlarmTreePV pv1, final AlarmTreePV pv2)
+                {
+                    final int cmp = pv1.getCurrentMessage().compareTo(pv2.getCurrentMessage());
+                    if (cmp != 0)
+                        return cmp;
+                    return super.doCompare(pv1, pv2);
+                }
+            };
         case DESCRIPTION:
         	return new AlarmComparator(up)
 			{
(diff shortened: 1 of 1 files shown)


kasemir:
You are introducing a new alarm comparator.
What you do in there puts alarms in this order:

//          1. Invalid / Disconnected alarm
//          2. Major unacknowledged alarm
//          3. Minor unacknowledged alarm
//          4. Alarm cleared but unacknowledged
//          5. Major acknowledged alarm
//          6. Minor acknowledged alarm

That's very close to the order provided by the SeverityLevel. ordinal(), see javadoc for https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/alarm/alarm-plugins/org.csstudio.alarm.beast/src/org/csstudio/alarm/beast/SeverityLevel.java

The order of alarm severities is used all over the alarm system, and it's defined in the SeverityLevel. The table column sorting code should not introduce a different order.

jbobnar:
> That's very close to the order provided by the SeverityLevel. ordinal()

It's close, but not the same. The difference is number 4. This is according to the specs I got from Nadine. If you disagree with the proposal, please close this request and we'll wait until everything is cleared up.

> The order of alarm severities is used all over the alarm system, and it's defined in the SeverityLevel.

That sorting is for the Icon column. It's not for the severity or current severity.

kasemir:
// 4. Alarm cleared but unacknowledged

Putting all alarms with current severity OK but unacknowledged latched severity on one level is not sufficient. Need to still sort on the unacknowledged, i.e. latched alarm severity.

Compared to everywhere else in the alarm system, I would sort by the latched alarm severity, and if they match, use the current alarm severity.
That puts them into a predictable order.

If ITER considers a current severity of "OK" as: Alarm cleared, all's fine, latched severity no longer matters, maybe you should simply configure your alarms to not latch?

--> Looks like this is also a topic for discussion, since we seem to disagree on the very basic sorting of alarm severities.

jbobnar:
> Putting all alarms with current severity OK but unacknowledged latched severity on 
> one level is not sufficient. Need to still sort on the unacknowledged, i.e. latched alarm severity.

If you look carefully, you'll see that if the two items that are compared are both cleared & not acknowledged, they will be compared by the latched severity (the else part of the if clause). The actual order is predictable and will be like this:
1. invalid unack active 
2. major unack active
3. minor unack active
4. invalid unack, cleared current severity
5. major unack, cleared current severity
6. minor unack, cleared current severity
7. major ack
8. minor ack
If two items are equal according this, they will be compared by super.doCompare, therefore compared by time and name.

> Compared to everywhere else in the alarm system, I would sort by the latched alarm 
> severity, and if they match, use the current alarm severity.

That would scatter the cleared alarms around (4.,5.,and 6. would be combined with 1.,2.,and 3.), which is not aligned with ITER requirements. The way I see it any alarm that has been cleared is less important than even a minor alarm that is still active.

If the above order is not acceptable for you, I suggest we make the sorting pluggable (through an extension point). The default implementation does what you propose, while ITER can implement the extension point to provide custom sorting.

kasemir:
> The way I see it any alarm that has been cleared is less important than
> even a minor alarm that is still active.

In the design of the alarm system, when you choose to latch alarms, that latched alarm severity matters. Each alarm has been carefully rationalized. A system expert came up with the motivation for the alarm, an IOC engineer identified or created a suitable trigger PV, an operator reviewed or updated the description, guidance, related display link.
When that alarm now triggers, it matters. Operators need to inspect and handle it.
If you have alarms that tend to come in briefly, then clear on their own after a short time, that's a nuisance alarm trigger that needs to be reconsidered.
If you decide that this is still a valid alarm, yet it can't be perfected, so it tends to trigger for a short while, then usually clears on its own, and that's fine, no further operator action required, it should be configured with a trigger delay and/or be non-latching.

utzeln:
To be discussed today, but as mentioned by Jaka, we want the latched behavior as if an abnormal event occurs the operator has to be informed.
But ITER operation considers that the operator should focus first on the non recovered alarms - invalid, major and minor - and then to the ones that have recovered. Perhaps they are wrong and in your experience the recovered ones help to analyse the non latched ones?

kasemir:
Notes from online discussion:

Can we solve this by adding a sorting method to the SeverityLevel that sorts on the combination of latched and active alarms?

The only difference: Need a preference to determine
a) Primary sort on latched, secondary on current, third on alarm time
b) Primary sort on current, secondary on latched, third on alarm time
?

If so:
 * Add such a sorting method to the SeverityLevel 
 * Alarm table's icon column uses that

At this time, not clear. ITER needs are not directly matched by b), it's a more complex table.

--> Need to see the complete table of severity level combinations

utzeln:
For ITER, we have the alarm severity and the alarm state.
The alarm states considered are:
• Acknowledged – when the operator has indicated awareness of the presence
of an alarm
• Unacknowledged – when the alarm is raised but the operator has not
indicated awareness of its presence
• Cleared – when the condition that triggered the alarm has returned to
normal
• Shelved – the operator has prevented a nuisance alarm from being displayed
for a limited time.
• Suppressed – alarm is suppressed when logical criteria are applied to
determine that the alarm shall not occur, even though the base alarm
condition is present.

So for ITER the primary sorting criteria is the alarm state - 1/unacknowledged, 2/cleared and 3/acknowledged. The secondary criteria is the severity - 1/invalid, 2/major and 3/minor.

In the current implementation, do we manage such alarm state independently to the alarm severity?

jbobnar:
> Can we solve this by adding a sorting method to the SeverityLevel that sorts on the combination of latched and active alarms?

This would not solve anything and would also be a bad design solution. We are sorting AlarmTreePV objects. SeverityLevel is just a criterion. It is an enum and should not depend on the AlarmTreePV.

Why can't we have sorting configurable through an extension point?

kasemir:
Nadine's description is helpful:
--------------
Acknowledged
Unacknowledged
Cleared
Shelved
Suppressed

Primary sorting criteria - 1/unacknowledged, 2/cleared and 3/acknowledged.
Secondary criteria is the severity - 1/invalid, 2/major and 3/minor.
--------------

You bring up new states that we haven't mentioned before. In the current alarm configuration, there is only 'disabled', not shelved nor suppressed.
Are you shelving alarms by disabling them?
Is suppression handled by alarm filters?
That might be yet another discussion.

For current PV severities, I think we all agree on the order of undefined, invalid, major, minor, ok as defined by EPICS.

The latched alarm severities can also be acknowledged, the current order is undefined, invalid, major, minor, undefined-ack, invalid-ack, major-ack, minor-ack, ok. Is there a fault in this order?

Sorting by alarm time obviously sorts by time, and sorting by PV name, description is alphabetical.

Most columns use a simple primary/secondary sort with these basic orderings.
Severity column: Sort by latched severity, then time stamp.
Current severity Column: Sort by current severity, then time stamp.
PV Name: Sort by PV name..

Your desired sort order for the icon column is basically by latched alarm severity, except that anything with current severity 'ok' is interleaved between active (unacknowledged) and acknowledged alarms.
I'm unsure about this. Why ignore the primary sorting criteria for a specific value of the secondary? To me, this seems to break the fundamentals of sorting, as if I had a telephone book that's sorted alphabetically, except people with a certain secondary attribute are placed in the middle, ignoring their name.

I tried to come up with another way to get your desired order, for example by using separate tables for active and acknowledged alarms, and then having different sort orders in each, but I can't.
Does this mean that the basic sort orders are wrong?

I assume you want 1/unacknowledged, 2/cleared because an alarm that already cleared is not as important as for example an active minor alarm. For the alarm system design, that's not really true. If an alarm was configured to latch, that was done because even a brief alarm matters. Otherwise there can be a time filter, or a non-latching trigger PV to avoid the alarm-but-currently-ok scenario.
When there are many alarms, operators might want to check the current severity. In the original table separation into active and ack'ed alarms, they'd then click on "Current Severity" to sort by that, maybe ack those that are 'OK', and go back to sorting by "Alarm Severity". At other times you may want to check what happened most recently, so you sort by time. Still, each column sort remains a basic primary/secondary sort.

utzeln:
You bring up new states that we haven't mentioned before. In the current alarm configuration, there is only 'disabled', not shelved nor suppressed.
Are you shelving alarms by disabling them? 
>> Yes, shelving means disabling for beast
Is suppression handled by alarm filters? 
>> Yes, the logic criteria to suppress the alarm will be the Enabling filter in beast.
That might be yet another discussion.
>> Yes, I mentioned it as we will also need at some point to setup an alarm table for shelved and suppressed alarms (since these alarms may constitute relevant status information for the operator). This means that even when the alarm is shelved or suppressed, the model maintains its priority updated.

berryma4:
@kasemir @utzeln @jbobnar what's the status here?

berryma4:
@kasemir @utzeln @jbobnar what's the status here?
Close, or move to 4.2.0?

kasemir:
On June 9, I had emailed this:

---
I think we need another hangout.

On one hand, I'm ready to give up and say: You can sort the icon column however you want. It's just one optional column. On the other hand, that sort doesn't make sense to me, because as I explained on the ticket it's like sorting names alphabetically, except every bald person is placed in the middle, regardless of name. Add to that using the title of the first guidance as an "ID" and its detail as an "Action", I'm afraid this changes the alarm table from a pretty straight forward generic tool into one very specific to ITER. Looks like you've come to the same conclusion, you're using an ITER copy of the alarm system, and are simply offering to merge the changes back.

Good news is that several of the changes are useful to all: Switching between separate & combined table, configurable columns, multiple table views with filter.
So I'd hate to see a complete split. At the same time, I also don't want to just merge any change. We did with the Data Browser, where we ended up with a dysfunctional pile of features that in the end didn't work together.

I'll be on travel from Friday on. Can we find a time before, like Wednesday the 10th, usual time?
---

There was no response.
I still think this icon sorting order makes no sense.

utzeln:
In order to make our June release, we needed to have a temporary local copy of the alarm table sort.

But, ITER strategy was and still is to collaborate and contribute to the shared effort and it is nice to see that some of our changes are welcome.

Regarding the alarm states and sort, we disagree - that is life.

Now, our main priority is BOY more than the alarm system. Nevertheless, I expect final requirements for the alarm system from Operation and the ergonomics study during fall this year. Then, I will need to asses how the current system is far from / close to ISA and industrial standards. As usual, I will try to convince everyone that the current system is the best one.

So the conclusion from my side is that it was a good lesson learned and even though the "ITER" alarm sort was integrated to our June release, it won't be used for production. So we can drop the change.

berryma4:
Sorry to take the thread over for a different topic.  But, @utzeln you
mentioned ISA.  This is great.  I believe if we can point to a standard, we
can have less disagreements.

utzeln:
As for now, regarding the sorting part, the ergonomics study source is "EEMUA and industry best practice".
Then, we have also "ISO 11064 recommends redundant coding, hence the use of colours and symbols.". 
And the use of flashing is based on IEC 62241. 

It seems that the final requirements for the alarm system will be based on ANSI/ISA 18.2 Alarm Management Standard.

jbobnar:
Rejected. Closing the pull request.

PR #1118, opened 2015-05-12
#818 fixing the repository poms to include the profile to upadte remote

shroffk:
Adding the Profile to push to remote repo


berryma4:
#818

Issue #1055, opened 2015-04-10
cs-studio eol and code format conventions


berryma4:
Taking this discussion to a issue ticket (from #1054 "818 1043 removing the non existant plugin from application.all.feature" ( d794b2261cda368008d14e8d0ee5670a475a77fb ))

kunal:
```
Oh no, I have the git.autocrlf enabled and eclipse configured to ignore whitespace....which is why this file was converted to UNIX endings and was committed without me noticing.

Fixing the all files now would make the history and blame useless...so we should preserve the line endings as they are and agree that all future files should be created with UNIX line endings.

We could add these git attributes to the cs-studio repository itself and thus ensure that this policy is followed...especially by careless people like me.
Additionally we should also define the java code format
currently the repository is a mix of 
1. eclipse convention (tab size=4)
2. java convention (tab size=8)
3. java convention with no tabs (only spaces)
This makes code really difficult to read at times.

@berryma4 @kasemir 
Should this be moved to a different issues?
```

richard:
```
At Diamond we didn't use core.autocrlf, preferring instead to set the line.separator property in each and every plugin.

See for example: https://github.com/openGDA/gda-core/blob/master/uk.ac.gda.core/.settings/org.eclipse.core.runtime.prefs

This has the advantage of enforcing the 'correct' setting, without relying on:

Manually setting core.autocrlf after cloning a repo.
Manually setting up the Eclipse workspace 'correctly'. (If each plugin specifies the 'correct' setting, the workspace setting doesn't matter.)
(Yes, this does mean repeating the same settings in each and every plugin; but it's trivial to check consistency across plugins with a script.)
```

kunal:

```
We can avoid having to manually setting core.autocrlf after cloning a repo.
I was considering adding .gitattributes file to the cs-studio repo, so there would be no need to manually set aore.autocrlf
https://help.github.com/articles/dealing-with-line-endings/

```

kasemir:
Yes, it might help a little if all contributors had similar Eclipse editor settings:

* Use Unix line endings (http://stackoverflow.com/questions/1886185/eclipse-and-windows-newlines)
* Use spaces instead of tabs (which is in the basic editor settings plus Java/Code Style/Formatter)
* Configure Java/Editor/Save Actions
  1) "Organize imports"
  2) "Additional Actions": Remove trailing white space on all lines

We're unlikely to agree on all those options.

I think we can agree on the basic whitespace handling.
[...]

PR #1141, opened 2015-05-14
Checkstyle

willrogers:
The checkstyle task will be executed with `-Pcheckstyle` for each of the module builds.


berryma4:
#1055

PR #1152, opened 2015-05-14
Svg animation updates

jbobnar:
Bugfixes and improvements of the SVG animation.


kasemir:
Updated MANIFEST.MF versions, but didn't update the corresponding pom.xml versions..
I think we fixed it.

PR #1167, opened 2015-05-19
LinkingContainerBugfix

jbobnar:
linking container edit part incorrectly set the opi path of its internal display model, which could result in unresolved relative paths.
This was broken by cf2b6a6. @berryma4 , please check if it still works for you.


berryma4:
You may also notice a pv connection issue in the linking container (multiplying connections on refresh that don't get dropped on close).

jbobnar:
We should rename the widget to "LeakingContainer". I think I got the connections under control, at least it closes whatever it has created. However, the jython library leaks like a sieve (and might keep a reference to a PV). It appears to be a known issue and they claim to have fixed some of the leaks in 2.7. Final version was released this month. Might be worth upgrading.

shroffk:
Ok, 
I was able to test that the pv readers and writers (using the opi screens under Issues in opi-regression-test repo) are not leaking however there are still update events that are lost.
This is a monster to debug and test

berryma4:
Yes, we should move to jython 2.7 in 4.2.x.  That, and I'm hoping JyNI project pans out.

PR #1168, opened 2015-05-20
Add simplepv.pvmanager into opibuilder feature.

willrogers:
Otherwise pvmanager is not recognised by OPIBuilder.

@kasemir @shroffk is this the correct feature to add it into?

### DIFF
--- a/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
+++ b/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
@@ -160,6 +160,13 @@ For details, see http://www.eclipse.org/legal/epl-v10.html
          unpack="false"/>
 
    <plugin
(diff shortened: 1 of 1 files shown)


berryma4:
I like the idea of this being in the main opibuilder feature

shroffk:
The simplepv is currently not used by any other applications, so it sort of makes sense to include this in the opibuilder.

We have 2 implementations simplepv.pvmanager and simplepv.vtypepv...should they have their own features?

berryma4:
That's why I like it.  This give preference to pvmanager.

kasemir:
I'm with Kunal: Only used by the opibuilder, so add both simplepv.pvmanager and simplepv.vtypepv to an opibuilder feature.

There's also org.csstudio.simplepv.utilitypv, which we should _not_ add, because the older utilitypv doesn't handle all data types and results in added conversions between DBR_* and VType.

shroffk:
@berryma4 can we merge this to proceed with testing

PR #1172, opened 2015-05-26
#818 adding the missing plugin to channelfinder.feature

shroffk:


berryma4:
@shroffk can you make this a clean merge

shroffk:
The build fails at a plugin which was not modified in this pull request.
Additionally I have updated the jenkins job to use jdk8 instead of 7.
can you trigger this build again

berryma4:
It doesn't merge clean

Issue #1113, opened 2015-05-08
 Filtering the content of the alarm table based on a (selected) node in the tree


berryma4:
- Not implemented yet, up for discussion
  - More of a UI design question 
  - Multiple instances of the alarm table, each showing alarms belonging to a particular node
    - related to filtering

kasemir:
I'm for allowing multiple alarm table instances, and for adding a filter to the alarm table.

 * Filter must be very obvious. For example, text at top of alarm table that shows: "Active Filter: /Some Area/System1/Level2"
 * Filtering must not be an accidental side effect of clicking around in the alarm tree. What would work is dragging a node in the alarm tree into the table -> Table now filters on that node and everything below. When using multiple alarm tables, this also makes it obvious which one will update its filter, i.e. the one into which the node is dropped.
 * Can filtering also be configured directly in the alarm table, for example by simply entering the filter path as "/Some Area/System1/Level2"?

utzeln:
For now, when that alarm table is split into Current Alarms and Acknowledged Alarms we have the number of alarms / lines shown.

Our initial proposal was to display for instance:
Current Alarms (20 of 56 alarms on /Some Area/System1/Level2)
20 being the number of alarms from the specified node
56 being the total number of alarms from root
 /Some Area/System1/Level2 being the node path used as filter

When grouped, we could have:
Current Alarms (20 of 56) Acknowledged Alarms (13) on /Some Area/System1/Level2

In the View menu, we also could add an option Show -> Show all to reset the filter to the root node:
![image](https://cloud.githubusercontent.com/assets/4686181/7678176/576d0140-fd52-11e4-9677-b270e286d92e.png)

This option Show, could be used to specify the path: Show -> Filter.

utzeln:
Finally, how to open a new alarm table view?

We could use the view menu and add an option "New Alarm Table"?

jbobnar:
To enable drag/drop of alarm tree nodes, the class org.csstudio.alarm.beast.TreeItem has to implement java.io.Serializable and some of the referenced classes need to do that as well. The AlarmClientModel field can be transient, but GDCDataStructure, AADataStructure, and Timestamp should implement Serializable.
The problem is that org.epics.util.time.Timestamp is out of our control (it is part of the PVManager). We can make it transient for now, but it should be fixed eventually. 

[...]

Issue #1112, opened 2015-05-08
Alarm Configure individual columns width, weight and title


berryma4:
- Not implemented, but it is agreed on as okay (preference w/o gui okay)
  - Kunal mentioned minimum width preference (lesson learned from logbook table / debian bug)
  - Kunal will verify 10 pixel minimum, check if after making width zero it can be re-sized again.

shroffk:
For the vanishing table columns on Debian7 and Debian6 Kay's fix works. I verified with the databrowser property table and the alarmtable. The minumum width of 10 pixels is enforced - if the column is minimized beyond this it jumps back out to 10 pixels.

We can consider this aspect resolved and move on to implementing the preferences for allowing the configuration of width/weight/title

Issue #1109, opened 2015-05-08
Alarm icons: Flash


kasemir:
Especially if active and acknowledged alarms are combined into one alarm table, it is helpful to have the active alarms stand out.
An option for flashing icons would provide this.

utzeln:
Detailed requirements: the alarm priority symbol/icon shall flash at 2Hz until acknowledged by the operator. The flash timing of symbols must be synchronized for all flashing states. 
The alarm message must never flash.

Remark: on the mimics also, it is required that the component’s symbol (valve, pump, breaker...) shall flash at 2Hz until acknowledged by the operator. Same "look & feel" for the GUI regardless if it is boy or beast.

PR #1188, opened 2015-06-08
Alarm Table Updates

jbobnar:
Implementation of the changes in the alarm table. Please review and if OK, I'll updated the help page and changelog. 

The major changes are: 
- multiple alarms table allowed (each can connect to a different configuration root): #1113 
- drag and drop items from the tree to the table to set a filter item: #1113 
- additional icons in the toolbar, to apply filter, select the entire root or show the tree root: #1113
- It is possible to configure the width, weight, title, and order or columns in the preferences: #1112 
- changing the width, order etc. in runtime will be preserved (for all tables): #1112 
- blinking icons of unacknowledged alarms (can be configured only via preferences) #1109 
- acknowledged and unacknowledged tables are synchronized (is the width of a column is changed or the columns are moved around, the other tables is updated accordingly)
- Added tooltips to all column headers
- Reorganized the alarm perspective to use perspective extensions instead of creating the layout programmatically.


berryma4:
@kasemir can you check for merge.

kasemir:
Yes.

jbobnar:
Implemented blinking synchronisation across all tables.
Filtering of alarms happens on a non GUI thread.
Updated changelog.

berryma4:
Can I close #1113, #1112, #1109 ?

PR #1192, opened 2015-06-09
Linking Container Bugfix

jbobnar:
Another bug was discovered by @utzeln : when linking containers were stacked the $(DID) marco was not properly initialized. That resulted in scripts being triggered by incorrect pvs. It seems this also goes back to #913.


berryma4:
thank you

PR #1194, opened 2015-06-10
Opi validation updates

jbobnar:
Added the opivalidation plugin to the opibuilder feature. Is that OK? Otherwise I need to add an additional feature, containing just this plugin.

Created an application and product configuration to run the validation in a headless mode from command line.

Fixed a few bugs.

### DIFF
--- a/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
+++ b/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
@@ -180,4 +180,11 @@ For details, see http://www.eclipse.org/legal/epl-v10.html
          version="0.0.0"
          unpack="false"/>
 
+  <plugin
+         id="org.csstudio.opibuilder.validation"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
 </feature>
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.validation/opivalidation.product
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.validation/opivalidation.product
@@ -7,8 +7,6 @@
    </configIni>
 
    <launcherArgs>
-      <programArgs>-rules validationRules.def -schema IOSchema.opi -opilocation g:/resources -results g:/temp.log
-      </programArgs>
       <vmArgsMac>-XstartOnFirstThread -Dorg.eclipse.swt.internal.carbon.smallFonts
       </vmArgsMac>
    </launcherArgs>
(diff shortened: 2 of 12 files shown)


berryma4:
@kasemir @shroffk do we want a new feature for this?  I want this, but does everyone?

shroffk:
Do you know the update site from which the org.eclipse.wst.validation dependency can be resolved? I think the target file might need to be updated to include this dependency.

kasemir:
It'd be nice for the target to be somewhat stable. Maybe change no more than once a month? Eclipse 4.5 Mars is to be released June 24. We should adopt that soon afterwards on 'master', which would also require a new target file.

jbobnar:
@shroffk wst.validation It's part of the eclipse update site.

The current opibuilder feature doesn't look like a proper feature, but more a collection of plugins that are related to opibuilder: adlParser, adl2boy and converter are probably also not needed at every installation.

I'd recommend to remove those plugins from the feature, which are not in the opibuilder-plugins and put them into another feature. The opi builder feature should then be configured to depend on that other feature. If a certain plugin is only used by opibuilder (and there is little or no chance that it will be used somewhere else) it should become a part of this repository.
The problem with the current setup is that if there was another feature that provides the same plugin, you will only be able to install both features in CSS if they both provide the exact same version of the shared plugin.

kasemir:
Which Eclipse update site? Kunal may be correct that it's not in our current cs-studio/target-files/target-dev.target that points to http://download.eclipse.org/eclipse/updates/4.4/R-4.4.2-201502041700.

The current opibuilder feature is certainly a collection of things that not everybody needs, but at the same time adlParser, adl2boy, EDM converter don't cause any visible harm to end users. They hardly add new menu entries, they don't affect the runtime at all. Similarly, I'd be OK with adding the validator. What might be useful is a split into an opiruntime feature (just the runtime) and an opieditor feature (editor, adl2boy, EDM converter, validation, ..).

shroffk:
@jbobnar 
1. I agree with the feature re-organization you mentioned. adlPraser, adl2boy and converter etc should be moved to a different feature but they should still be part of the opibuilder module since they are closely dependent on opibuilder.

2. For the validator I am fine with it being in the main opibuilder feature, so long and the default def file is completely commented out.

3. @kasemir just the runtime could be very interesting.

jbobnar:
@kasemir @shroffk 

Sorry, it's the eclipse releases site, not the update site: http://download.eclipse.org/releases/luna/

berryma4:
@kasemir @shroffk @jbobnar at this point, does this go on 4.1.0 or 4.2.0.  @utzeln do you need this on 4.1.0?

berryma4:
I do like the idea of a opiruntime feature (just the runtime) and an opieditor feature, but this will have to happen in 4.2.x.  I'm merging this to 4.1.x.

Issue #1197, opened 2015-06-10
upgrading to jython 2.7 has significant impact on script loading times


shroffk:
With the move from jython 2.5 to 2.7 loading python scripts takes a very large amount of time
I have added an opi screen in the regression testing repo
https://github.com/ControlSystemStudio/opi-regression-tests/tree/master/Issues/Issue_1197

With jython 2.5 this screen loads in 5-7 seconds
With jython 2.7 this screen loads in 100-115 seconds

berryma4:
http://www.jython.org/latest.html#jython-2-7b3-release-notes
By default, site module is imported when using PythonInterpreter Use -Dpython.import.site=false (as of RC1) to not import site at all

shroffk:
Reverting back to 2.5.3 resolves this porblme but this requires code changes since the PythonInterpreter has changed and no longer has the close() method

kasemir:
Reverting back is not an option for us because that Jython version is out of date.
For example, it cannot handle catching exceptions "e as …", so the PyScanClient won't work.

kasemir:
-Dpython.import.site=false makes no difference when I try that here. I'm afraid your example screen nicely shows why scripts should be limited to exceptional cases, can't be used for each widget all the time. Also shows why a significant update in that area is needed, since that script loading currently blocks the whole UI.

berryma4:
Odd the VM argument made a difference for me

shroffk:
I completely agree...I already spoke to my users and showed them alternative ways of doing this.

But this is a regression, usually my users don't take kindly to being told that a lot of their old screens are now broken :)

I still feel at times formulas can be very useful (like in this case they don't want to create IOCs to control the color)

kasemir:
I ran the current CSS with your Widgets_with_python_scripts.opi in JProfiler, and it's indeed spending all the time in org.python.util.PythonInterpreter <init>. I get the same ~100 seconds as Kunal, even with -Dpython.import.site=false. I don't have the source code for org.python.jython_2.7.0.b3.jar, so can't look further inside what exactly is happening there.

What helps: The previous org.python_2.5.2.20140409 plugin was expanded, while org.python.jython_2.7.0.b3.jar is a jar file. I've unpacked it in the product's plugin dir, then updated configuration/org.eclipse.equinox.simpleconfigurator/bundles.info. Now the example loads in ~35 seconds.

So we need to package that python plugin in expanded form, as a directory, not jar.

[...]

PR #1198, opened 2015-06-11
Issue 1197 jython slow

berryma4:
#1197


berryma4:
@shroffk @kasemir @jbobnar 
Please check that this doesn't break your existing screens.
This upgrades to jython 2.7.0 and changes the new PythonInterpreter to threadLocalStateInterpreter in BOY.

kasemir:
I haven't rebuilt everything to get jython 2.7.0. I did check the opibuilder updates (threadLocalStateInterpreter) and it looks good. Kunal's very slow example is fixed, and the SNS displays that I checked are OK as well.

jbobnar:
I think there are two (potential) issues:
1. The synchronization in JythonInit#getInstance is not done properly. If two threads request the first instance at the same time, you'll end up doing the same initialization twice.
2. I didn't look into details, but Py.getSystemState seems like a shared instance among all interpreters on the same thread. If that is true it probably shouldn't be closed and cleaned (in dispose method)?

kasemir:
You're right on the init. Making the whole call synced would help. .. Except it is (unfortunately) called on the UI thread, so in practice no difference.

I understand the threadLocalStateInterpreter to return a new interpreter, 
http://www.jython.org/javadoc/org/python/util/PythonInterpreter.html#threadLocalStateInterpreter(org.python.core.PyObject)
Closing it seems OK.

jbobnar:
Perhaps the initialization could be done during plugin activation in a background thread?

kasemir:
Jaka, you're right about performing the init once, in background, on startup, if possible. Kunal, Eric and I had discussed the multiplicity a bit. Originally, each script invocation created its own interpreter and context for variables. With the jython update, the interpreter initialization time suddenly went way up, and one interpreter per script is really a lot. Then Eric went to just one interpreter for all. Much faster, but probably bad because now a script invoked in screen B would see vars created by another script in screen A. There could also be a problem with longer running scripts that spawned threads and access the 'widget' var, which is set to the current widget on each script invocation. Having one interpreter per display seems a good balance. This would even allow a script run on display startup to set variables and later running scripts within the same display can then access them. Would need to be aware of 'widget' changing, though. Way into the future, scripts should run in the background. Scripts for multiple widgets should run in parallel, so they need to implement something like a 'run(widget, pvs)' method that is invoked with the current parameters. Setting those as global vars won't work.

berryma4:
There is really no point to JythonInit with the current threadLocalStateInterpreter, so I removed it.  But, you're right, with this and pvmanager having the lazy load, the first screen takes a little long.

PR #1204, opened 2015-06-12
o.c.opibuilder: Update shared links when an OPI is launched.

willrogers:
In particular, this means after CS-Studio is started.  This requires
CS-Studio to understand a different syntax:

`cs-studio --launcher.openFile "/path/to/file macro1=value1¬links"`

Links are in the same format as the `-share_link` command-line argument.
The ¬ for separator may be changed before release.

I discussed this feature with @kasemir @shroffk @berryma4 at the code-a-thon.
See #544.  There are some gymnastics in keeping this off the UI thread but opening the OPI ON the UI thread.  You may want to review the threading.

It also allows opening OPI shells using the Position macro.


kasemir:
I'm OK with this except for the separator character that you already mention, because it's not on US keyboards.

Find some other character out of  ```~!@#$%&*-+```.

Or use "-share_link" as the separator?
```
--launcher.openFile "/path/to/file macro1=value1 -share_link /some/path/folder=/workspace1"
```

willrogers:
OK I'm back from Malawi now and I will look at this.  I think the `-share_link` option is best.

willrogers:
See #1204.

willrogers:
This now works with the `-share_link` option.  It's possible we might still want to refine this before the 4.2 release, but I think it basically works correctly.

Issue #1206, opened 2015-06-16
boy: symbol widgets slow over http


berryma4:
symbol widgets are loaded much too slow over http

berryma4:
I've added caching with google guava.
Although, the linking container maybe loading these images more than it did in the past.

PR #1207, opened 2015-06-16
o.c.swt.widgets: added caching for symbol images using google guava

berryma4:
#1206


berryma4:
@kasemir @shroffk @jbobnar @utzeln 
Could you please check these changes.
It only goes to the cache if it's not in the workspace.

shroffk:
1. The connection cleanup works, there are no more leaked reader and writers from the linking container.

Problems: 
1. referencing groups in linking containers is broken
2. the connections between widgets inside linking containers no longer works (but maybe this is ok)

kasemir:
Replaced tabs with spaces and updated the javadoc since pathToInputStream() no longer handles workspace files.

berryma4:
Thank you.

Issue #1190, opened 2015-06-09
Release 4.1.x (master)


berryma4:
Proposing to freeze the master branch to 4.1.x.

  - [x] resolve caj/pvmanager issue #1174
  - [x] create diag feature to prevent inclusion of duplication versions of applications like probe and pvtree
  - [x] python scripts take significantly longer to load #1197
  - [x] NPE for xyGraph #1200 #1199
  - [x] report BOY regression tests (ControlSystemStudio/opi-regression-tests)
     https://docs.google.com/forms/d/1cKILK51oWpVcy5iDCe4gPK6mvCSvlMWIS5JazuO2_Lc/viewform
  - [x] make 4.1.x branch
  - [x] make new github milestone 4.2.0 - testing (master)
  - [x] make new github milestone 4.1.1 - stable (4.1.x branch)
  - [x] make Jenkins csstudio.product release
  - [x] run Jenkins csstudio.product release

Site Representatives, are there any other outstanding issues that need to be addressed before release of 4.1.x?
@berryma4, @kasemir, @shroffk, @richardfearn, @willrogers, @utzeln, @mjclarke01, @jpenning, @ksdj

jpenning:
We're fine with this, keep up the good work :-)

Jörg.

On 09.06.2015 15:46, Eric Berryman wrote:
>
> Proposing to freeze the master branch to 4.1.x.
>
>   * resolve caj/pvmanager issue #1174
>     <https://github.com/ControlSystemStudio/cs-studio/issues/1174>
>   * report BOY regression tests
>     (ControlSystemStudio/opi-regression-tests)
>     https://docs.google.com/forms/d/1cKILK51oWpVcy5iDCe4gPK6mvCSvlMWIS5JazuO2_Lc/viewform
>
>   * make 4.1.x branch
>   * make new github milestone 4.2.0 - testing (master)
>   * make new github milestone 4.1.1 - stable (4.1.x branch)
>   * make / run Jenkins csstudio.product release
>
> Site Representatives, are there any other outstanding issues that need 
> to be addressed before release of 4.1.x?
> @berryma4 <https://github.com/berryma4>, @kasemir 
> <https://github.com/kasemir>, @shroffk <https://github.com/shroffk>, 
> @richardfearn <https://github.com/richardfearn>, @willrogers 
> <https://github.com/willrogers>, @utzeln <https://github.com/utzeln>, 
[...]

Issue #1103, opened 2015-05-06
Connection from Linking container


shroffk:
The modification to the linking container made in the master branch there seem to be significant error with handling connections.

I have attached opi screens to re create some of the problems.
https://github.com/ControlSystemStudio/opi-regression-tests/tree/master/Issues/Issue_1103

1. The linking container + simplepv.PVManagerPV is leaking references to pv readers and writers
Open the Datasource debugger (CS-Studio --> debugging --> Datasources)
Run the Parent_connection_simple_loc.opi a couple of time, you will notice that the number of (T)otal number of listeners does not return to 0 when all opi screen are closed and no loner in the run mode.

2a. Readers are not correctly attached and Events are lost
Run the Parent_connection_simple_loc.opi, some of the text update widgets will have the correctly initialized value 0, refreshing the opi will show that some text update widgets will randomly render the correct value. 

2b. Attempts are made to write to PV's before the are created.
```
2015-05-06 10:17:55.415 WARNING [Thread 61] org.csstudio.simplepv.pvmanager.PVManagerPV (stop) - PV loc://SR_C25(0) has already been stopped or was not started yet.
2015-05-06 10:17:55.419 WARNING [Thread 61] org.csstudio.simplepv.pvmanager.PVManagerPV (stop) - PV loc://SR_C25 has already been stopped or was not started yet.
```

kasemir:
Likely caused by #912: Linking container is broken, requiring additional `if (widget instance of LinkingContainer)` type checks when starting/stopping PVs for all widgets.

jbobnar:
2a and 2b are bugs in PVManager and PVManagerPV. They are not limited to the LinkingContainer. They are visible also if you use the child opi directly, just less frequently. Hit the refresh button like crazy and you're bound to see it happen.

2b was partially fixed by #1167. The org.csstudio.simplepv.pvmanager has incorrect "synchronization" of start/stop methods. From the git comments I see some deadlocks were being fixed this way, so I do not dare to change this. But here are my thoughts:
- PVManager#start and PVManager#stop should be synchronized, because they are not thread safe. There is a (small) window when both methods could be executed synchronously. Luckily they are always called from the same thread, so a problem like that cannot happen.
- Shouldn't the stop method return after it puts a "stopping" runnable on the notification thread?
- If the notification thread is not a single thread executor, strange things might happen.

[...]

PR #1208, opened 2015-06-17
Issue1103 revert linkingcontainer

berryma4:
reverting work from #912 that caused pv connection issues.
Mainly, the issue is from the add/remove of children in XMLUtil.fillLinkingContainer and the add/remove of children in configureDisplayModel overlap property fires sometimes.
Side note, adding/removing children seems like a bad way to refresh the model anyway.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
@@ -44,6 +44,7 @@ import org.eclipse.gef.EditPart;
 import org.eclipse.gef.EditPolicy;
 import org.eclipse.gef.GraphicalViewer;
 import org.eclipse.gef.editparts.ZoomListener;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IActionFilter;
 
 /**The Editpart Controller for a linking Container
@@ -207,6 +208,7 @@ public class LinkingContainerEditpart extends AbstractLinkingContainerEditpart{
                 widgetModel.setDisplayModelOpiRuntime(widgetModel.getRootDisplayModel(false).getOpiRuntime());
             }
         });
+        
 
         connectionList = displayModel.getConnectionList();
         if(connectionList !=null && !connectionList.isEmpty()){
@@ -255,6 +257,15 @@ public class LinkingContainerEditpart extends AbstractLinkingContainerEditpart{
 
         LinkedHashMap<String,String> map = new LinkedHashMap<>();
         AbstractContainerModel loadTarget = displayModel;
+        
+        if(!widgetModel.getGroupName().trim().equals("")){ //$NON-NLS-1$
+            AbstractWidgetModel group =
+                displayModel.getChildByName(widgetModel.getGroupName());
+            if(group != null && group instanceof AbstractContainerModel){
+                loadTarget = (AbstractContainerModel) group;
+            }
+        }
+        
         // Load "LCID" macro whose value is unique to this instance of Linking Container.
         if (widgetModel.getExecutionMode() == ExecutionMode.RUN_MODE) {
             map.put("LCID", "LCID_" + getLinkingContainerID());
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
@@ -191,9 +191,18 @@ public class LinkingContainerEditpart extends AbstractLinkingContainerEditpart{
         //This need to be executed after GUI created.
         if(getWidgetModel().getDisplayModel() == null) {
             IPath path = getWidgetModel().getOPIFilePath();
-            getWidgetModel().setDisplayModel(new DisplayModel(path));
+           
+            final DisplayModel tempDisplayModel = new DisplayModel(path);
+            getWidgetModel().setDisplayModel(tempDisplayModel);
+            try {
+                XMLUtil.fillDisplayModelFromInputStream(
+                        ResourceUtil.pathToInputStream(path), tempDisplayModel,
+                        getViewer().getControl().getDisplay());
+            } catch (Exception e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
+            }
         }
-
 
         LinkingContainerModel widgetModel = getWidgetModel();
         DisplayModel displayModel = widgetModel.getDisplayModel();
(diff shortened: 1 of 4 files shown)


berryma4:
#1103

berryma4:
@shroffk @kasemir @jbobnar @utzeln

shroffk:
Solves: 
1. The connection cleanup works, there are no more leaked reader and writers from the linking container.

Problems: 
1. referencing groups in linking containers is broken
2. the connections between widgets inside linking containers no longer works (but maybe this is ok)

kasemir:
Something is broken with the script lookup.
When I have a display loaded from http://path/to/file.opi and it references a script in a relative location scripts/example.py. i.e.  http://path/to/scripts/example.py

Looks like a repeat of #998 , except this time there isn't any error message, just nothing happens.

kasemir:
The connection cleanup is indeed better, and the connections between widgets inside linking containers is simply not possible because the content it meant to be anonymous.

Need more time to look into the script issue.

kasemir:
Good news: The scripts run just fine.

The problem is in the resizing of the linking container, as can also be seen in 
BOY Examples/Miscellaneous/DynamicLoadWidgets/LoadWidgetsExample.opi

That code adds containers like this:
```
var linkingContainer = 
 WidgetUtil.createWidgetModel(
     "org.csstudio.opibuilder.widgets.linkingContainer");	
linkingContainer.setPropertyValue("opi_file", "SubPanel.opi");

// 1 == "Size the container to fit the linked *.opi"
linkingContainer.setPropertyValue("resize_behaviour", 1);
```

On 'master', that works OK.
On this branch, the linking containers all have zero width and height, they don't size to match the SubPanel.opi.

kasemir:
As far as I understand, this is happening:

Script calls ```widget.addChildToBottom(linkingContainer);```

This ends in ```LinkingContainerEditPart.configureDisplayModel()``` line 297 calling ```performAutosize()```.
At that time, the linking container has no child widgets, so it resizes to 0.

On the master branch it works OK because the linking container does have child widgets at that point (which are then removed and loaded again?).

The fix might be to invoke ```performAutosize()``` later, after the child widgets are loaded, once and for all.

kasemir:
Seems like this update results in never loading the "SubPanel.opi" at all.
The script sets the "opi_file" property of the linking container, but the linking container never loads it, and the resize then shrinks it because it's empty.

@berryma4 Eric, I'm not sure how to fix this since I'm not familiar with the code.
The _editor_ code related to the linking container seems to load the opi_file, but the runtime which would have to do this when the opi_file property is changed at runtime no longer does.
The 4_2_LinkingContainer.opi example has an "Example 1: Dynamic Symbol" based on a Rule which also fails to update since again the runtime changes to the opi_file property are ignored.

If loading the content in the "editor", then unloading, then loading again in the "runtime" was the original problem, I think it'd be OK to only load in the runtime. So in the editor you'd no longer see the content in the linking container. Pity, but most important is that it shows up in the runtime.

Or check the mode: In run mode, load only in the "runtime". In edit mode, also load the content in the "editor".

berryma4:
Yes, when the displaymodel is null, it wasn't handled correct.

berryma4:
I need to update the error handling, but I want to make sure this works for you too.

kasemir:
BOY Examples/Miscellaneous/DynamicLoadWidgets as well as our SNS script work after your displaymodel null update.

The 4_2_LinkingContainer.opi "Example 1: Dynamic Symbol" is still static.
It updates the "group_name" property to show only a selected group within a linked file.
I'm not sure if that's an essential feature, because again the content of a linking container should be anonymous, you can't run 'connectors' to anything inside, so why should you know about groups inside?

berryma4:
Well, the Example 3 on 4_2_LinkingContainer.opi, with the "thumbnails" is not working for me.  I'm guessing the auto size is off.

kasemir:
Yes, the "thumbnails" are broken as well. That's since the change  in LinkingContainerModel when the boolean "zoom_to_fit" and "auto_size" properties were replaced by one enumerated "resize_behaviour" property.
It's somewhat backward compatible, but not 100%.

berryma4:
I see it has the resize behavior of "Size the container to fit the linked *.opi".
So, it's acting correct, but property is wrong.  It should be "size *.opi to fit the container".

berryma4:
Is there anymore we can do to help the backward compat?

berryma4:
LinkingContainerModel: Line 169
``` java
  public void processVersionDifference(org.osgi.framework.Version boyVersionOnFile) {
        super.processVersionDifference(boyVersionOnFile);
        if(boyVersionOnFile.compareTo(VERSION_CHANGE_OF_RESIZE_BEHAVIOUR) < 0) {
            Activator.getLogger().log(Level.CONFIG, "Converting linking container to new style of resizing behaviour.");
            if((Boolean)getPropertyValue(PROP_AUTO_SIZE)) {
                setPropertyValue(PROP_RESIZE_BEHAVIOUR, ResizeBehaviour.SIZE_CONTAINER_TO_OPI.ordinal());
            } else if((Boolean)getPropertyValue(PROP_ZOOMTOFITALL)) {
                setPropertyValue(PROP_RESIZE_BEHAVIOUR, ResizeBehaviour.SIZE_OPI_TO_CONTAINER.ordinal());
            } else {
                setPropertyValue(PROP_RESIZE_BEHAVIOUR, ResizeBehaviour.SCROLL_OPI.ordinal());
            }
        }
    };
```

The thumbnail has:
```xml
    <auto_size>false</auto_size>
    <scripts />
    <zoom_to_fit>true</zoom_to_fit>
    <height>149</height>
    <name>Linking Container</name>
```

The dynamic triangle has:
```xml
    <auto_size>false</auto_size>
    <scripts />
    <zoom_to_fit>false</zoom_to_fit>
    <height>49</height>
    <name>Linking Container_1</name>
```

berryma4:
Okay, I understand.  It modifies when you open in edit mode, not runtime.

berryma4:
Strange, the defaults seem to be used instead of the file values. 
```java
    protected void configureProperties() {

        addProperty(new BooleanProperty(PROP_ZOOMTOFITALL, "Zoom to Fit", WidgetPropertyCategory.Display, true));
        setPropertyVisibleAndSavable(PROP_ZOOMTOFITALL, false, false);

        addProperty(new BooleanProperty(PROP_AUTO_SIZE, "Auto Size", WidgetPropertyCategory.Display, true));
        setPropertyVisibleAndSavable(PROP_AUTO_SIZE, false, false);

        addProperty(new ComboProperty(PROP_RESIZE_BEHAVIOUR, "Resize Behaviour",
                WidgetPropertyCategory.Display, ResizeBehaviour.stringValues, ResizeBehaviour.SIZE_OPI_TO_CONTAINER.ordinal()));
    }
```

kasemir:
Is processVersionDifference() simply not invoked?

berryma4:
It's invoked, it's just that the property values are the defaults instead of the xmlfile values

kasemir:
I think we should merge as is.

On master, the LinkingContainer is broken. Our nightly build is useless. This pull fixes the connection issues, it "works".

The thumbnail issue in 4_2_LinkingContainer.opi, the translation of old "zoom_to_fit" and "auto_size" properties into the new "resize_behaviour" property is the same on master and with these changes.
This pull is not introducing that as a new bug, so it could become its own ticket, to be addressed separately, without delaying the more essential make-it-work-at-all fix.

shroffk:
Let us merge this and create the 4.1 release now.
Bug fixes can continue

berryma4:
I removed the processVersionDifference for the LinkingContainerModel.  It seems to work better without it.  Does it still work for everyone?

kasemir:
Removing the processVersionDifference seems counter-intuitive, but sure enough the 4_2_LinkingContainer.opi example works better that way.
May not even need a ticket until loading certain older files turns into a real big problem.

berryma4:
allows release #1190

PR #1212, opened 2015-06-24
PV Table comments and row insert/move

### LINKED ISSUE #844: PV Table "Description" column
SNS plans to use some larger PV Table configs as a convenient way to edit, save, compare, restore a bunch of device settings, because it's easier to auto-gen a *.pvs file than a more elborate *.opi display.

They would benefit from having a 'Description' column that goes with each PV, not just the plain PV name. By default, that description would be read from the *.DESC PV, i.e. a field of the same record.
But the config file would allow specifying 
 a) No description PV at all -> column remains empty
 b) Totally different PV, like some other String record
 c) By default, it'd use the main PV of that table line + ".DESC"

There'll be a preference setting to disable the "Description" column, i.e. get the current behavior.
Configuring the description PV for each line will only be possible in the *.pvs file format, not when loading/saving *.sav autosave files.

kasemir:
This closes #844:

PV Names that start with "#" are treated as comments.
This also allows adding empty rows as "#".

Context menu to add a new row anywhere (not just appending at the end).
Drag-drop can now also be used to move existing rows around.

Sorry that there was already a pull request yesterday with which this could have been combined.
Didn't expect to find time to close #844 so soon.


berryma4:
... just in time.  I was running the release script for 4.1.0 release, ... and stopped it.
Let's add this to the new 4.1.x branch, I will merge it over.

kasemir:
Thank you!!

Users here had asked for the 'description' 6 month ago. That was a simple thing, but overall the development got stuck. So I'm very happy that it's now available combined with the comments etc.

Issue #1219, opened 2015-06-25
Linking container fails to reload display model when macros are changed


shroffk:
You can recreate the problem with the following example screen
/BOY Examples/Miscellaneous/DynamicMacros/DynamicMacros.opi

kasemir:
I'll create a pull request to get this onto master

PR #1220, opened 2015-06-25
#1219 compare the macromap and recreate the display model if there is a difference

shroffk:
When a linking container is being reloaded, there is an additional check to see if the macro map of the display model matches the widget model...if it does not then the display model is discarded

@berryma4 @kasemir can you check


berryma4:
#1219

PR #1224, opened 2015-07-01
Two minor bugfixes in csstudio perspective and databrowser

jbobnar:
CSStudioPerspective: setting the layout via preferences did not work due to incorrect parsing of parameters.
DataBrowser was throwing ArrayIndexOutOfBoundsExceptions when the chart was empty.

### DIFF
--- a/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/model/Model.java
+++ b/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/model/Model.java
@@ -236,16 +236,23 @@ public class Model
         return axes.size();
     }
 
-    /** Get specific axis
+    /** Get specific axis. If the axis for the specified index does not exist, method returns null.
      *
      *  <p>Thread-safe access to multiple axes should use <code>getAxes()</code>
      *
      *  @param index Axis index
-     *  @return {@link AxisConfig}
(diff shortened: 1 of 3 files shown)


kasemir:
Can you create a ticket for the data browser ArrayIndexOOB issue?

Because of ".. event may come from .. background thread", the pattern
```
if (index < model.getAxisCount())
 {
   final AxisConfig axis = model.getAxis(index);
```
fails when the axes change concurrently.

Something like this would be better:
```
final AxisConfig axis = model.getAxis(index);
if (axis == null)
    return; // Axis no longer valid
```
Might require changes to the model, though, to have getAxis() be thread safe and return null when the index is no longer valid.

jbobnar:
This would not work 
```
final AxisConfig axis = model.getAxis(index);
if (axis == null) ...
```
because the first line (to get axis from the model) already throws an exception. The model needs to be adapted slightly. I'll make an issue and push a commit.

kasemir:
Right, it does require a change to model.getAxis(index).
In there, please avoid synchronize because the underlying array is already a CopyOnWriteArrayList and adding synchronize in one place will then cause FindBugs to require sync. in every place.
Instead, use something like
```
try
{
   return axes.get(index);
}
catch (ArrayIndexOutOfBounds ex)
{
   return null;
}
```

jbobnar:
Bugfix: Issue  #1226

kasemir:
I'm fine with the data browser change, thanks for that.
On the perspective-from-preferences update, Kunal added that, but as far as I can tell the update is good.

Issue #1226, opened 2015-07-01
ArrayIndexOutOfBoundsException in Databrowser


jbobnar:
In case of an empty databrowser chart an ArrayIndexOutOfBoundsException is thrown when axis is updated.

org.csstudio.logging.PluginLogListener (logging) - Unhandled event loop exception
java.lang.ArrayIndexOutOfBoundsException: 0
       at java.util.concurrent.CopyOnWriteArrayList.get(CopyOnWriteArrayList.java:387)
       at java.util.concurrent.CopyOnWriteArrayList.get(CopyOnWriteArrayList.java:396)
       at org.csstudio.trends.databrowser2.model.Model.getAxis(Model.java:248)
       at org.csstudio.trends.databrowser2.ui.Controller$3.valueAxisChanged(Controller.java:263)
       at org.csstudio.trends.databrowser2.ui.ModelBasedPlot$2.lambda$1(ModelBasedPlot.java:105)
       at org.csstudio.trends.databrowser2.ui.ModelBasedPlot$2$$Lambda$39/117556243.accept(Unknown
Source)
       at java.util.Optional.ifPresent(Optional.java:159)
       at org.csstudio.trends.databrowser2.ui.ModelBasedPlot$2.changedYAxis(ModelBasedPlot.java:105)
       at org.csstudio.swt.rtplot.internal.Plot.fireYAxisChange(Plot.java:1160)
       at org.csstudio.swt.rtplot.internal.Plot.mouseUp(Plot.java:1056)
       at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:220)
       at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
       at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4454)
...

PR #1229, opened 2015-07-06
Bugfix

berryma4:


berryma4:
#1226 bugfix, moving to 4.1.x from master

PR #1238, opened 2015-07-10
opibuilder: Fix ExecutePythonScriptAction NPE and use separate interp.

### LINKED ISSUE #1237: Jython Script problems
The changes to fix Jython memory leaks and re-use the Jython interpreter/state introduced two new problems.

```
java.lang.NullPointerException
        at org.csstudio.opibuilder.widgetActions.ExecutePythonScriptAction.dispose(ExecutePythonScriptAction.java:169)
        at org.csstudio.opibuilder.editparts.AbstractBaseEditPart.deactivate(AbstractBaseEditPart.java:302)
````
because it's trying to access the member variable `state` which is no longer set.

PyScanClient scripts fail when opening a display within the same CSS instance for the second time because something inside the url2lib or socket lib is closed, yet still reused because it's the same Jython interpreter/state: https://github.com/PythonScanClient/PyScanClient/issues/18

### PR DESCRIPTION
**Pull request by kasemir**
This reverts back from the threadLocalStateInterpreter to a PythonInterpreter for script actions invoked by action buttons.
The threadLocalStateInterpreter leaves the python PyScanClient/urllib2/socket lib in a state that results in https://github.com/PythonScanClient/PyScanClient/issues/18

This reversal of the PythonInterpreter -> threadLocalStateInterpreter commits only affects script actions invoked by action buttons, not display scripts triggered by PVs.

In addition, it fixes a bug introduced by commit e95684b which tries to clean the 'state' but found that as `null`.

Fixes #1237

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/widgetActions/ExecutePythonScriptAction.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/widgetActions/ExecutePythonScriptAction.java
@@ -27,7 +27,6 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.gef.GraphicalViewer;
 import org.eclipse.swt.widgets.Display;
-import org.python.core.Py;
 import org.python.core.PyCode;
 import org.python.core.PyObject;
 import org.python.core.PyString;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by berryma4** [comment]
Does this make it slow again?

**Comment by kasemir** [comment]
Note that this only applies to action buttons, not scripts triggered by PVs.

Yes, the first time you press an Action Button which invokes a script, there is the time to create the jython interpreter. Press the button again -> Fast. Close that *.opi, re-open, press the button for the first time -> Loads jython interpreter.

But I don't see an alternative. With the interpreter being re-used, I cannot get the python urllib2 to work.
Basic code like this:
```
            req = urllib2.Request(url)
            response = opener.open(req)
            result = response.read()
            response.close()
            # Actually just 'pass', but call it anyway
            req.close()
```

only runs once. Second invocation results in the error shown in 
https://github.com/PythonScanClient/PyScanClient/issues/18, an error inside the jython implementation of _socket.py

We use scripts on action buttons with the scan library for beam line automation.
If that doesn't work, we cannot use CSS on SNS beam lines.

kasemir:
I actually peeked into `org.python.jython_2.7.0/Lib/_socket.py`, tries calling its `_shutdown_threadpool()` to enforce cleanup so the next time the script runs it starts over fresh, but didn't find anything that works other than starting a new jython interpreter as we used to do.

PR #1239, opened 2015-07-13
o.c.o.converter.test: remove explicit JRE from .classpath

willrogers:
The new plugin specified an explicit JRE container.


berryma4:
thank you!

PR #1244, opened 2015-07-14
Data Browser 'units'

### LINKED ISSUE #1233: Data Browser Annotations: Include/allow units. Select one or give better error.
Annotations should have a way to include units.
Maybe make that another replacement item, and then use {0} {1} {2} {3} as the default, with 0, .., 3 being name, position, value and (new) units.

When adding an annotation without selecting any of the traces in the "Add Annotation" dialog, the dialog stays up and an array index error is logged. Should give an error message like "Please select a trace". In addition, a trace should be pre-selected, which seems to happen on Linux but not Mac OS X in the current version.

kasemir:
Fixes #1233 : Annotations, value axes, legend, .. show the channel's units.

Fixes a legend bug (didn't reset color).

Follows Kunal's suggestion to use PlotListener interface with defaults to eliminate need for PlotListenerAdapter. .. had to rename it to RTPlotListener to avoid name clash with same class in Data Browser.

Tried to update RAP version of RTPlot.


berryma4:
Can I cherry pick these two commits, or are they dependent on the changes in master?

kasemir:
Looks like you can.

There is one difference in the data browser on master compared to 4.2:
The DB on master no longer automatically switches to the DB Perspective from the toolbar button, because automatically switching perspectives would mess us up when we use persps. for arranging *.opi files.

PR #1288, opened 2015-08-04
Add uid & plugin configuration to AlarmConfigTool.product

richardfearn:
(done to AlarmServer.product in 0b971ec)

### DIFF
--- a/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/AlarmConfigTool.product
+++ b/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/AlarmConfigTool.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="AlarmConfigTool" id="org.csstudio.alarm.beast.configtool.product" application="org.csstudio.alarm.beast.configtool.application" version="4.0.0.qualifier" useFeatures="false" includeLaunchers="true">
+<product name="AlarmConfigTool" uid="alarm-config" id="org.csstudio.alarm.beast.configtool.product" application="org.csstudio.alarm.beast.configtool.application" version="4.0.0.qualifier" useFeatures="false" includeLaunchers="true">
 
    <configIni use="default">
    </configIni>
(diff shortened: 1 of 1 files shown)


kasemir:
How are you actually using this?

Since we moved to maven/tycho, I thought the *.product needs to be in a repository/ subdir, so the products I actually use are in https://github.com/ControlSystemStudio/org.csstudio.sns/tree/master/repository/*.product

I thought about deleting the *.product files that used to be in primary plugin of each product, but left them in place for historical reasons, since they could still be used for launching in the IDE .. until the dependencies get out of sync too far.

richardfearn:
> How are you actually using this?

I copy it to a {{repository}} module, like the one used by ITER and SNS.

> I thought about deleting the *.product files that used to be in primary plugin of each product

But then where would the definitive product definitions for things like the alarm server or alarm config tool be?

I wondered if they should go into the {{org.csstudio.product}} repo, since there's just a CS-Studio GUI product in there at the moment.

Ideally I wouldn't need to copy them at all, but you can't build products from them in their current locations. Actually I don't know if there are any site-specific tweaks that might be necessary, even for things like the alarm server or alarm config tool, that might necessitate a site-specific product definition.

kasemir:
So far, my only site specific tweaks are to the CSS GUI product.

These are all generic:
 * alarm config
 * alarm notifier
 * alarm server
 * archive config
 * archive engine
 * jms2rdb
 * scan server

Again, the *.product files that I actually use happen to be in https://github.com/ControlSystemStudio/org.csstudio.sns/tree/master/repository

They are used in production. By now they likely differ from the original *.product files because I had to add plugin start level settings and update the list of plugins, just as you did with the  alarm config tool.product.

I agree with having https://github.com/ControlSystemStudio/org.csstudio.product/tree/master/repository contain all these product files, and removing them from the original locations in the cs-studio repo.

Meaning: Remove cs-studio/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/AlarmConfigTool.product and instead put that into org.csstudio.product/repository.
Maybe instead of moving the original ones, which as you just saw with the alarm config tool may already be out of date, use a copy from org.csstudio.sns/repository.

Do you have time to do that?

berryma4:
instead, could we make new repos for:
 - org.csstudio.alarm: config, notifier, server, jms2rdb
 - org.cssstudio.archive: config, engine
 - org.csstudioscanserver: scan server

I'm currently making these for our site this week (including debian
packages), maybe I should make them more generic and put them on
github/controlsystemstudio?

richardfearn:
> I'm currently making these for our site this week (including debian packages), maybe I should make them more generic and put them on github/controlsystemstudio?

Yes please! At present there's no easy way for a new site to build anything besides the CS-Studio GUI. A set of `.product` files inside an `eclipse-repository` module would be very useful.

richardfearn:
> I'm currently making these for our site this week (including debian packages), maybe I should make them more generic and put them on github/controlsystemstudio?

@berryma4 how are you getting on with these?

Issue #1270, opened 2015-07-27
databrowser: archive appliance falls back to raw data


berryma4:
For some reason, databrowser keeps falling back to raw data (not optimized), which cs-studio kills because that is way too many points (see attached dump). @kasemir @jbobnar @shroffk 
![image](https://cloud.githubusercontent.com/assets/854021/8908485/58840db0-344a-11e5-8736-30ee2abbf7a9.png)

This is a single pv from 1/1/2015 to today.  
Also, in "Inspect Samples", I don't see statistics.

jbobnar:
I just tried this locally and it works fine.

Which version of appliance are you using? Do you have a chance to step through with a debugger? If yes, put a breakpoint on line 142 in ApplianceArchiveReader and let us know what happens there.

berryma4:
I'm updating to pbrawclient 0.0.4, it seems to fix the issue.

berryma4:
This did not completely fix the issue

berryma4:
strange with this in my preferences file:
org.csstudio.archive.reader.appliance/useStatisticsForOptimizedData=true
ApplianceArchiveReader in line 148, useStatistics is sometimes false?

Now they are working?
But, the stats look wrong:
![image](https://cloud.githubusercontent.com/assets/854021/8915969/4d9451ae-3476-11e5-87ab-41940d1dda1f.png)

berryma4:
the problem may be with the archive appliance ... (other than the useStatics sometimes false).
When ApplianceArchiveReader:142 sometimes comes in as zero, then fails into the catch which is a raw value iterator.

I'm not sure if I like this failing over to raw.  This will kill most clients pretty fast.

berryma4:
After removing from the preferences, I can't seem to reproduce this.
Although the stats are strange:
![image](https://cloud.githubusercontent.com/assets/854021/8916310/a6501ccc-3478-11e5-824c-efb4bdc05fc5.png)

berryma4:
All I can think was maybe the end of line in the preferences file was messed up?

berryma4:
End of lines were the same, I'm not sure what's going on here.
![image](https://cloud.githubusercontent.com/assets/854021/8916706/5890a940-347b-11e5-840e-54c29b0febfd.png)

berryma4:
Also, ApplianceArchiveReader ln:142 sometimes returns 0
then uses the raw iterator.
I'm guessing this is a appliance issue?

[...]

PR #1291, opened 2015-08-05
Implemented the new optimized iterator for faster data retrieval of scalars

jbobnar:
Implemented the optimized data retrieval (related to discussion in #1270) and updated the pbrawclient jar.
There is a new setting in preference.ini: useNewOptimizedOperator.

It can be used to enable or disable the new optimized data retrieval. If the appliance installation is old and does not yet support the optimized retrieval, set this setting to false. If you have an upgraded installation, which supports the optimized retrieval, set this to true. If set to true and it fails, it will always fall back to the old mechanism.


berryma4:
thank you!

Issue #1293, opened 2015-08-10
Linked resources no longer work in BOY


jbobnar:
If a project has a linkedResource in .project file, the resource is not loaded, if OPI is started in standalone mode or as a view.

jbobnar:
Fixed.

jbobnar:
Projects imported into workspace are also not handled correctly. 
#1300 fixes that issue.

jbobnar:
Resolved.

PR #1302, opened 2015-08-18
More problems with OPI paths

jbobnar:
Another attempt to fix this: #1293 
I think I finally understand what was going on with the paths, so I hope this is the final pull request for this.

The new opi launching messed up a few things. It worked fine when the launched file was self sufficient in its own environment. But, immediately when the file needed to access something from the workspace, there were problems if the file was not physically located in workspace. The reason being that the launcher approach is primarily intended for launching external editors.

This pull request introduces a middle step in the launching procedure. It initializes an editor which launches the content and then closes automatically. It is a workaround but it is the only way I know how to keep the paths "workspace aware" without having to do too much file system magic.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/META-INF/MANIFEST.MF
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: RCP fragment of opibuilder
 Bundle-SymbolicName: org.csstudio.opibuilder.rcp;singleton:=true
-Bundle-Version: 3.2.2.qualifier
+Bundle-Version: 3.2.3.qualifier
 Bundle-Vendor: Xihui Chen<chenx1@ornl.gov> - SNS
 Fragment-Host: org.csstudio.opibuilder;bundle-version="2.0.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/pom.xml
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/pom.xml
@@ -8,6 +8,6 @@
     <version>0.0.1-SNAPSHOT</version>
   </parent>
   <artifactId>org.csstudio.opibuilder.rcp</artifactId>
-  <version>3.2.2-SNAPSHOT</version>
+  <version>3.2.3-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 </project>
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/src/org/csstudio/opibuilder/runmode/DisplayLauncher.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/src/org/csstudio/opibuilder/runmode/DisplayLauncher.java
@@ -10,7 +10,6 @@ package org.csstudio.opibuilder.runmode;
 import java.util.Optional;
 
 import org.csstudio.opibuilder.runmode.RunModeService.DisplayMode;
-import org.csstudio.opibuilder.util.ResourceUtilSSHelperImpl;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.ui.IEditorLauncher;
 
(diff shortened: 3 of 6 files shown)


berryma4:
@kasemir @shroffk I'm not sure about this work around, is there another way?
Does this affect opis over http?

kasemir:
Loading of *.opi files, related displays and scripts via http:// used to be OK.
What Jaka wants to address, I think, is a problem with linked folders. When you have those, you can open the file in a text editor or display editor, but opening in the display runtime failed.

Not sure if the temporary RCP editor part is the best solution, but haven't looked that that myself to come up with anything better, so maybe merge this for now?

willrogers:
I can confirm that the previous commit broke links for us.  This commit appears to fix them again.  The implementation is somewhat involved.

I see that the temporary editor part does appear then disappear, which is not ideal.  Is there a way to make it invisible?   This might be useful for another reason.

willrogers:
You might be able to hide the part, but `hideEditor()` is not supported in e4: https://bugs.eclipse.org/bugs/show_bug.cgi?id=374132

I think means that it wouldn't be supported by the compatibility layer.  It doesn't seem to work for me.

kasemir:
To make sure I understand the problem:

My workspace might be `/home/freddy/MyCSSWorkspace`.
In my workspace, I have a link `displays` to `/usr/local/displays`.

When I now try to open one of the displays from the navigator, the launcher is invoked with the absolute file system location, for example `/usr/local/displays/example.opi`.

The code that's currently in `org.csstudio.opibuilder.runmode.DisplayLauncher` fails to convert the absolute file location back into a workspace file. We could update that code, somewhat like this:
Loop over all files in the workspace, get their 'location', and compare to the absolute file path. Since you can have multiple linked folders, there would be more than one workspace file for a given absolute path.
The methods `IWorkspaceRoot.findFilesForLocation*` do that, and have the same caveat in their javadoc.

Not sure what's better:
a) Locate the first possible workspace file, use that. .. and it may be the wrong one.
b) Use an editor that receives the exact workspace file, but then we really don't want to show that editor.

I'm leaning towards trying to locate the file. If there are multiple matches, that could be a warning, since it's usually not ideal if your workspace lists the same files multiple times.

kasemir:
For example, this "works" for files in linked Folders:

```
public class DisplayLauncher implements IEditorLauncher
{
    @Override
    public void open(final IPath path)
    {
        final IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
        final IFile[] files = root.findFilesForLocation(path);
        RunModeService.openDisplay(files[0].getFullPath(), Optional.empty(), DisplayMode.NEW_TAB, Optional.empty());
    }
}
```

Needs check for files == null and files.length > 1.

jbobnar:
Kay's understanding is correct. The solution proposed above would work in most cases, but as noticed, the problem is if the file is linked from more than one location. This solution seems to cover those situations as well: 

```
ISelectionService service = PlatformUI.getWorkbench().getActiveWorkbenchWindow().getSelectionService();
ISelection selection = service.getSelection();
if (selection instanceof IStructuredSelection) {
    Iterator elements = ((IStructuredSelection)selection).iterator();
    while(elements.hasNext()) {
        Object e = elements.next();
        if (e instanceof IResource && ((IResource)e).getLocation().equals(path)) {
            OPIShell.openOPIShell(((IResource)e).getFullPath(), null);
            return;
        }
    }
}
```

kasemir:
Using the currently selected resource, presumably from the Navigator where the context menu was invoked, seems to go in the back and out the ear, but I think you've found the best solution!!

No hack to open a temporary EditPart, no possibly lengthy search for matching workspace files that could run into duplicates.

When you add the selection-based fix, please include comments to describe why we're using the selection instead of the `IPath` that's already provided to `IEditorLauncher#open()`.
Might also make sense to check, somewhat like this:
````
public class DisplayLauncher implements IEditorLauncher
{
    @Override
    public void open(final IPath path)
    {
        // path is absolute file location, but we need a workspace resource.
        // Since we're invoked from the Navigator, the current selection should provide
        // the desired path
  
        ISelectionService service = ..
         ..
 
        IResource selected_resouce = ..
    
        // Check that path from selection indeed matches the launcher input
        // TODO Just 'equals' may not be good enough for this test.
        if (selected_resouce.getFullPath().equals(path))
             RunModeService.openDisplay(selected_resouce.getFullPath(), ..);
        else
        {
           // Log error:
           // The current selection is a file in the workspace,
           // but its location in the file system doesn't match what we are supposed to open
        }
    }
}

````

jbobnar:
Done. 

There is one issue remaining, but I think it is so unlikely to happen that we can ignore it. See here:
https://github.com/css-iter/cs-studio/commit/163b9b9854c3ee01d9e052ab4d57e6bbcffa5862#diff-e77a0159383d90e0afc6e3f334a10feaR52

kasemir:
This should work.
Thanks!

berryma4:
Thank you everyone.

PR #1315, opened 2015-08-27
Various minor bugs fixed

jbobnar:
Fixed two minor bugs in the alarm table: activate the alarm table view when opened, check if table is busy before making an update.

Removed the jmsmonitor plugin from debugging feature, because the plugin was moved to another feature.

Fixed a regular expression when parsing the RGB info for console.


kasemir:
Thanks for figuring out what to do about the "Viewer is busy" messages!

Issue #1318, opened 2015-08-27
Creation of shared links fail if any do not exist 


nickbattam:
Launching CS-Studio with share-links specified on the command line fails if any of the linked locations do not exist. An exception is thrown and none of the specified links are created.

It would be helpful if this behaviour can be changed so that a Warning log message is recorded for 'broken' links, while the valid links are created as expected.

PR #1319, opened 2015-08-27
Fail gracefully if LinkedResource does not exist

nickbattam:
As discussed in #1318


kasemir:
Good idea.

However, instead of

````
log.warning(ex.getMessage());
````

I would suggest
````
log.log(Level.WARNING, "Cannot create link " + folder_path, ex);
````

The latter will log the complete stack dump, so you know why it failed.
The former could actually not print anything because ex.getMessage() can be empty.

kasemir:
Thanks!

Issue #1326, opened 2015-09-02
If you drag an OPIView out of the main window, it is repositioned


willrogers:
There is logic in `OPIView.java` to reposition a view if it has been opened from the navigator.

However, if you drag a view out of the main window, it resizes and also attempts to move to the location specified in the OPI file.

The resizing may be correct behaviour.  However, we have users that say that if you drag a view out then it should stay where it was put!  I agree.

willrogers:
@jbobnar I expect that you may need to make some changes after this commit to manage RCP and RAP.

Any comments on the changes are welcome.

kasemir:
Can this be closed since the pull request fixed it?

willrogers:
Yes, this is done.

PR #1327, opened 2015-09-02
Attempt at better logic for detached views.

willrogers:
* Resize whenever a view is detached, to make the view the same
  size as the OPI file
* Locate to place specified in OPI file only if programmatically
  detached.

Don't merge this yet!  It needs review from @kasemir @shroffk @berryma4 etc.

As discussed in #1326.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/runmode/OPIView.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/runmode/OPIView.java
@@ -89,8 +89,6 @@ public class OPIView extends ViewPart implements IOPIRuntime
 
     private static boolean openFromPerspective = false;
 
-    private static boolean openedByUser = false;
-
     public OPIView()
     {
         opiRuntimeDelegate = new OPIRuntimeDelegate(this);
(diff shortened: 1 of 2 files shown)


berryma4:
Comments?  @kasemir @shroffk

kasemir:
I think it's a good idea.

berryma4:
me too.

shroffk:
Agreed, ok for merge

kasemir:
One case where this doesn't work if you have two views within the same detached window:

Drag one view out of the main workbench to get a detached view,
then drag another view into that same detached view.
Arrange to be side-by-side, adjust window size.
--> One detached window with 2 views in it, side by side.

When you now restart CSS, each of those detached view resizes the overall detached window to just the size of that view
--> One window with those 2 views side by side, but the overall window is too small because it has the size of the view that was restored last and thus set the overall window size.

willrogers:
This behaviour hasn't changed due to my changes - it was like that before.

I agree that this feels wrong.  The user I'm talking to about this thinks that the original Eclipse behaviour is actually fine.  What happens there is when you drag the view out it remains the size and shape it was in the IDE.  However, when you resize and save a perspective (or restart Eclipse) the detached view stays where it was.

Previously someone has added some logic to make the views appear at a certain size and shape.  I don't mind if we remove this logic completely.  What do you think?

willrogers:
Actually, that's not true.  If we remove those lines then **when opened from a button** the view doesn't honour its size and location, and that's when we want it to (I assume).

It is difficult to know whether a view is newly created and detached, or detached because it's just been dragged out of a window.

willrogers:
I think f2332c51abe25f0ba18c4674f7284cfbdabdb8cb may be closer to what we want.  

* locates and resizes detached view if opened from a button
* leaves locations and size the same if dragged from the workbench

That is, unless I've missed something...

kasemir:
I think you found the lines which cause the problem, and the crux is indeed that we can't tell if that view was just opened from a button, or manually detached by dragging it out of a workbench window, or restored after an overall restart.

I tried removing those lines, and it fixes the misguided resize attempts on restart.
When dragging out, I think we're also better off without those lines because the view then drags out "as is", either keeping its size when you drop it freely, or keeping the size of the existing window into which it was dropped. Pretty intuitive, no magic size changes.

So when opening a "Detached Tab" view from a button is really the only case where resizing would make sense. I think the best location to handle this would be inside the RunModeService.openDisplayInView(): That's the place that opens the view which knows that it's detached. RunModeService line 233ff even has the code for it:

```
// Adjust position
if (position == Position.DETACHED) {
        SingleSourcePlugin.getUIHelper().detachView(opiView);
        opiView.positionFromModel();
}
```

That looks like it's meant to do exactly what we want. For some reason it's not doing it, though.

kasemir:
Sorry, looks like you already figured out what's wrong in `opiView.positionFromModel()` while I was writing the last comment.

Yes, your f2332c51abe25f0ba18c4674f7284cfbdabdb8cb is it!

Issue #1337, opened 2015-09-10
o.c.utility.file: creates file in executing directory


berryma4:
![wqer](https://cloud.githubusercontent.com/assets/854021/9798506/572877fa-57cf-11e5-85a1-9fceafd80c83.png)

It seems for URLs, o.c.utility.file creates a file in the executing directory, then makes an ifile from that in the workspace.

So, sometimes the directory the file is created is not writable.

berryma4:
.... this whole URL, and files / workspace in general should probably be rethought into a decent utility.

PR #1338, opened 2015-09-10
o.c.utility.file: creates file in executing directory #1337

berryma4:
I think there is a larger issue, but this fixes the problem.
Basically, don't make a file, use the inputstream to make the iFile.


berryma4:
@shroffk @kasemir

kasemir:
OK.

berryma4:
#1337

Issue #1335, opened 2015-09-10
Minor problems with EDM conversion


willrogers:
Conversion problems:

* PV name conversion produces invalid results since assertions are used and the `-ea` command-line flag is not generally used
* EDM and CSS treat empty macros differently
* Changing colour of 2D profile monitor is not supported
* 'Toggle' and 'push' behaviour of BooleanButton (EDM Button) is incorrect

Test problems:

* Using both junit3 and junit4 features in `EdmConverterTest` appeared to cause some problems on Jenkins, even though local Maven builds worked.
* Tests fail because:
 * Converted widgets are now one pixel bigger
 * The structure of the `EdmColor` class has changed

willrogers:
While I'm on this, I'll move all the tests to JUnit 4.

berryma4:
can we close this?

willrogers:
Yes, thanks.

PR #1341, opened 2015-09-11
o.c.o.converter: several changes to conversion and unit tests

willrogers:
The hybrid between junit3 and junit4 caused occasional build problems.


willrogers:
See also #1335

Issue #1354, opened 2015-09-25
OPIShell should not show FullScreen and Compact Mode options


willrogers:
Full screen and compact mode options don't make sense in the OPI shell context menu.

berryma4:
can we close?

willrogers:
Should be fixed.

PR #1355, opened 2015-09-25
o.c.opibuilder: edit context menu items

willrogers:
Only show 'full screen' and 'compact mode' menu items for
OPIView, not OPIShell.

As discussed in #1354.  @kasemir @utzeln 

This is an ugly method and not a pretty part of OPIBuilder in general, but I believe this does what we want.


kasemir:
Thanks!

Issue #1360, opened 2015-09-29
OPIShells may lose right-click menu


willrogers:
Because of the way Eclipse RCP works, each OPI Shell requires to be 'registered' against an existing View or Editor for the right-click menu to work.  At present, if the original view is closed, the menu no longer appears.

My proposed fix is this: create a view that summarises the open OPI shells.  Right-click menus can be attached to this view.  This view also allows you to find or close the OPI shells.  Potential problems:

* if you close this view, right-click menus can still be lost.  Reopening the view will restore them
* ideally, the view would not be closeable for this reason.  I can't figure out how to achieve this.

willrogers:
This ticket can be closed.

PR #1361, opened 2015-09-29
Opishell status

willrogers:
As discussed in #1360.  This requires review @berryma4 @kasemir @shroffk .


willrogers:
standalone-view

Here's what it looks like.

Issue #1171, opened 2015-05-22
Diirt integration


berryma4:
We would like to start using the newest diirt (3.x).

To do:
  - [x] fork diirt/diirt to controlsystemstudio/diirt
    - remote/diirt -> diirt ; origin/master -> dev ; origin/1.2_cs-studio -> releases
  - [x] update serviceloader to also use osgi service
  - [x] configuration / preferences
   - Add an "XMLPreferenceScope" that will read from for example a jca.xml config file that's compatible with webpods, but write changes from preference UI to InstanceScope
  - [x] make sure current Jenkins job works
    - fix p2 publish url and fix documentation
  - [x] update cs-studio to use new namespaces and api changes
  - [x] change plugins that used pvmanager extension points to use osgi services (custom formulas and datasources)

shroffk:
gabriele said:
```
Hi,

Relooking at the code, the best thing to do is transform the org.diirt.util.Configuration class into an interface. Have a DiirtHomeConfiguration, which is the default. CS-Studio can have a CSStudioCompatibilityConfiguration that takes the information from the current preferences. The product can then look at an eclipse preference “diirtConfigurationFromEclipseProperties”, and at startup change the default configuration.

That would be the way to go. That way is transparent to both the user and the rest of the diirt framework.

We should wait to see if the ITER pans out. If so, I’d do the interface work. It would not be that much.

Gabriele
```

shroffk:
Diirt integration issues

workspace projects can’t be used in the IDE - you always have the runtime (out of sync)

refreshing project, updating maven project have no effect.
They are in sync when first loaded into the workspace, but after the first build they are no longer.
Which is due to the fact that the manifest is produced in the build, this manifest timestamp does not match with the one the IDE has and viola you have the out of sync problem.

berryma4:
added branch: https://github.com/ControlSystemStudio/cs-studio/tree/diirt_integration

Most of this was taken from various attempts in the past.

migrated to diirt 3.0; 
o.c.u.pvmanager.ui, 
o.c.graphene.opiwidgets, 
o.c.graphene, 
o.c.channel.widgets, 
o.c.diag.epics.pvtree, 
o.c.diag.pvmanager.probe, 
o.c.opibuilder.widgets.extra, 
o.c.u.pvmanager.ui.toolbox, 
o.c.u.pvmanager.widgets, 
o.c.diag.pvmanager.javafx.probe

berryma4:
[...]

PR #1362, opened 2015-09-29
Diirt integration

shroffk:
Preliminary set of changes for diirt 3 integration
@arnaudf-iter @kasemir @willrogers @berryma4


shroffk:
In addition to the diirt 3 I have also tried to fix references to third part libraries like (javax, jdbc for mysql,..) to use the one provided by orbits.
In terms of dependencies, I have preferred the use of imported packages to required plug-ins to allow us some flexibility with the migration.

shroffk:
Note: There isn't a specific target definition file associated with this branch, the following chnages have been tested by building the diirt3 p2 repo locally and using it via the composite repository mvn property.

willrogers:
I find three problems when building:

* `org.csstudio.core.diirt.feature` can't find `datasource-timecache` for some reason - it's not getting built into the diirt p2 repo
* `org.csstudio.scan.server` still has references to `org.epics.vtype`
* `org.csstudio.simplepv.pvmanager.test` calls `CompositeDataSource.setDefaultDataSource()`, which no longer exists

Having said that, I still don't know how to configure OPIBuilder to connect to PVs when it is built.

willrogers:
That fixed problems one and two.  I still need to comment `org.csstudio.simplepv.pvmanager.test` to get Maven to build.

willrogers:
OK I built our product (including `org.csstudio.core.diirt.feature`) and followed the instructions [here](https://github.com/diirt/diirt/wiki/Diirt-Configuration) to configure it.  It works exactly as you'd expect.

shroffk:
Ok, that is some good news.
Did you try to change the configuration folder or are you using the default one which points to the user home directory.

TODO:
- [ ] Clean up the preference/property mapping for the diirt home folder.
- [ ] Put together a set of instructions about all the different ways we can configure diirt and what is the priority

willrogers:
I created a specific directory and used the preference to point to that directory.  I did get a couple of exceptions when the directory didn't exist.  I'm not sure if that's just teething problems.

berryma4:
#1171

Issue #1064, opened 2015-04-16
Update to Eclipse 4.5 Mars


kasemir:
Eclipse 4.5 aka Mars is expected end of June 2015.

CS-Studio 'master' should then be updated to successfully build against it.

Things to gain for CSS:
 *  Customize Perspective Dialog is back (ITER had missed it).
 * UI responsiveness monitoring. We should include & enable this, which will put information about UI freezes, for example code that accesses files on the UI thread, into the log, so we can then fix frequent problems.
 * org.eclipse.core.runtime.jobs.JobGroup helps with creating jobs that use all CPUs (can also be done with Executors.newWorkStealingPool() if connection to Progress View is not required)
 * SWT on Linux can use GTK+ 3 (I see org.eclipse.swt.internal.gtk.version=2.24.23 in Help, About, Installation Detail, Configuration on our Linux computers, so that GTK 2 is also still supported).
 *  JFace StructuredViewer has IStructuredSelection getStructuredSelection(), no more casts.
 * In addition to the plugin_customization.ini, products can have a *.pde file with saved preferences.
 * General/Appearance has a "Presentation" schema that could be useful when using CSS in a meeting on projector display, where right now the fonts are all too small
 * IAdaptable.getAdapter() now uses generics, i.e. returns the requested type without casts
 * Mac OS X products are proper *.app (https://bugs.eclipse.org/bugs/show_bug.cgi?id=431116)

Things to gain in the IDE:
 * Nested/Hierarchical view of projects that might better reflect how we now structure things in the repository
 * Target platform "Reload" may help those who use remote repositories
 * E4 model editor etc. now included in IDE, no longer a separate install
 * Model fragments can use xpath.

For more: https://www.eclipse.org/eclipse/news/4.5/M7, which ends in links to ../M6, M5, ..

berryma4:
I have also removed 32-bit builds

berryma4:
0.23.1 tycho?

Issue #788, opened 2014-11-04
Problems compiling CSS applications from master branch


ttrubep1:
Hi there,

I experienced some problems while trying to compile CSS from source and I hope someone here has an idea about what went wrong.

I tried to compile CSS from source using the instructions given in http://cs-studio.sourceforge.net/docbook/css_book.pdf.

I installed Apache Maven 3.2.3 and checked out ControlSystemStudio/maven-osgi-bundles@57ace426db3c094961defaf68df887d17a85f47b, ControlSystemStudio/cs-studio-thirdparty@254f468f2cc05c6391609479f0ce32b0027537ce and finally ControlSystemStudio/cs-studio@ffd1aac2a3f3bb3255343d32c367917575014a7c, all on branch master.

I compiled using `mvn clean verify`. This went well for maven-osgi-bundles, cs-studio-thirdparty and cs-studio/core. On cs-studio/applications, maven failed at the last step `applications-repository` and complained about missing dependencies.

The output of maven was the following:

```
[INFO] ------------------------------------------------------------------------                                                                                                                                                              
[INFO] Building applications-repository 4.0.0-SNAPSHOT                                                                                                                                                                                       
[INFO] ------------------------------------------------------------------------                                                                                                                                                              
[INFO]                                                                                                                                                                                                                                       
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ applications-repository ---                                                                                                                                                        
[INFO] Deleting /nfs/data/jollyj/tobias/slowcontrol/css-devel/cs-studio/applications/repository/target                                                                                                                                       
[INFO]                                                                                                                                                                                                                                       
[INFO] --- tycho-packaging-plugin:0.20.0:build-qualifier-aggregator (default-build-qualifier-aggregator) @ applications-repository ---                                                                                                       
[INFO] The project's OSGi version is 4.0.0.201411032218                                                                                                                                                                                      
[INFO]                                                                                                                                                                                                                                       
[INFO] --- maven-clean-plugin:2.5:clean (default-clean-1) @ applications-repository ---                                                                                                                                                      
[INFO]                                                                                                                                                                                                                                       
[INFO] --- jacoco-maven-plugin:0.5.3.201107060350:prepare-agent (default) @ applications-repository ---                                                                                                                                      
[...]

PR #1375, opened 2015-10-08
move to tycho 0.23.1, removed rap from build, o.c.diag.probe: fixed p…

berryma4:
…eriod_format error

#1064
@jbobnar I had to remove some rap features so it would compile.
I was getting this error again:
https://github.com/ControlSystemStudio/cs-studio/issues/788

### DIFF
--- a/applications/alarm/alarm-features/pom.xml
+++ b/applications/alarm/alarm-features/pom.xml
@@ -10,7 +10,6 @@
   <packaging>pom</packaging>
   <modules>
     <module>org.csstudio.alarm.beast.ui.feature</module>
-    <module>org.csstudio.alarm.beast.ui.rap.feature</module>
     <module>org.csstudio.alarm.beast.tools.feature</module>
     <module>org.csstudio.alarm.beast.ui.alarmtable.opiwidget.feature</module>
   </modules>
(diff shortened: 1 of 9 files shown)


jbobnar:
OK. I'll have a look at RAP (there are a few other things that need to be updated besides features).

berryma4:
Thank you!

jbobnar:
There is an "extra" folder in the cs-studio-thirdparty which contains rap.draw2d and rap.gef plugins (plus a few others): https://github.com/css-iter/cs-studio-thirdparty/tree/master/extras/plugins
All these plugins are obsolete and should not be used. However, tycho is eager to take the highest version number of any plugin it uses. For some reason the latest rap.gef and rap.draw2d have the version numbers 0.1.0. This results in using the extra plugins instead of the real latest ones.

Could you remove the entire extra folder from the thirdparty repository (or at least all rap plugins)?

kasemir:
Jaka, I think right now you know best how to build the RAP products. If you don't need those extra plugins, or on the contrary, they actually cause trouble, by all means remove them.

berryma4:
I removed the "extra" folder in thirdparty.  I will always be more than happy to remove anything from maven-osgi-bundles or thirdparty.

PR #1377, opened 2015-10-09
Diirt integration

shroffk:
Redoing the pull request again into master for the 4.2 release #1356

Please address this pull request before merging other pull requests since this touches quite a few files and accepting other pull request might render this one unacceptable.

Review:
@berryma4 @kasemir @willrogers @utzeln

Note:
Delete the remote branch after the merge is approved.

PR #1379, opened 2015-10-09
#1377 #1171 adding the new diirt p2 repository to the core and applications pom.xml

shroffk:
- Adds the missing diirt p2 repositories needs with the lastest move to diirt 3
- fixing the versions in the diirt module to 4.2


berryma4:
#1377 #1171

PR #1380, opened 2015-10-12
o.c.diag.probe: fix merge conflict

willrogers:
@berryma4 @shroffk just make sure I haven't missed something here.

See #1375,


kasemir:
I had been wondering about that. #1375 included "o.c.diag.probe: fixed period_format error", but as far as I could tell it only moved the initialization `period_format = NumberFormat.getNumberInstance(); `
from the constructor to the member variable. I actually think that's better, whatever you can initialize right away is preferred. But the previous init in the constructor was functional as well, and the change added another semicolon.

shroffk:
It seems like by diff was not correctly imported by Eric...thanks for fixing it @willrogers 

@kasemir, yup...the only change is to move the initialization out form the constructor. 
Since it is used in a final runnable in java 8_u60 this gives a compilation error. The eclipse IDE always complained but it used to compile fine with older versions of java 8 so we had just let it be.

Issue #1303, opened 2015-08-18
OPI Display in Standalone Window issues


utzeln:
Making some tests, I found the following issues with OPI opened in a standalone window:

1/ Full Screen - Compact Mode affect CS-Studio editor part not the runtime window

2/ no zoom and OPI history toolbar - when the OPI is run using the green button, there is a new toolbar which is nice. This toolbar is not available in the standalone window

3/ cannot create snapshot for a new logbook entry

kasemir:
I believe this is by design. That standalone window mode is supposed to be just like EDM: No toolbar, no history, no .. DLS opens those windows via the --launcher.openFile mechanism and arranges them on the screen via X11 tools.

For SNS, we are not interested in using that mode. We have looked at disabling the 'standalone' mode via a preference setting, but there was no easy way to do this because it's added just like the other "Open with ..." options via plugin.xml entries.

What might be possible is move those plugin.xml entries into opibuilder fragments, so each site can then decide if their product should include them, or not.

willrogers:
1) I agree that the context menu options for compact mode and full screen should not be there for OPI shells.

2) This may be less relevant for people using OPI shells.  Certainly it is not how we expect the OPI shells to work.  This might be the reason that you want to disable the OPI shell functionality.

3) I am not familiar with this.  Is it a missing context menu option?

willrogers:
The other problem you mentioned is that there is no longer an option to open an OPI in a whole new Eclipse window.  See #544.

PR #1392, opened 2015-10-15
Opi runtime/view switch

jbobnar:
Related to #1303 and #1333.
I made two fragments: one adds the view/standalone functionality and the other one uses the old opi runtime. You can use one or the other, but not both.
To use it, just add either the org.csstudio.opibuilder.rcp.opiruntime org org.csstudio.opibuilder.rcp.view to your product configuration. I already added the view fragment to the main opibuilder feature, so if your product is feature based, there is nothing you need to do to keep it the old ways.

While this solution works it is not the best design-wise. The OPIView still has to be part of the main plugin and so does the OPIRuntime (there is also RAP...). Ideally, the opibuilder should be redesigned to decouple these kind of things entirely, but that would be a much larger project. Since there is already plan to redo the opibuilder, I don't think it is worth doing it now, but should be kept in mind.


jbobnar:
> I would suggest, the need to have the new view behaviour fragment as part of 
> the boy feature, requiring someone to make their our feature if they want the old behaviour.

This is already done. The opibuilder-features/org.csstudio.applications.opibuilder.feature from this pull request contains the .view fragment. For anyone who is using that feature, no further actions are required.

I realize that the names of the fragment plugins are maybe not the most adequate. I am open to suggestions.

jbobnar:
Are there any plans to merge this pull request?

jbobnar:
Due to lack of interest to accept this change, ITER will try to solve this issue by patching the opibuilder. It needs to be done for the beast integration anyway.

Issue #1396, opened 2015-10-19
Pass parameters to commands in ScriptUtil


willrogers:
Can I squeeze one more feature in before the 4.2 freeze?

ScriptUtil allows you to execute arbitrary Eclipse commands.  

A small change would allow you to pass parameters as well to create parameterized commands.

berryma4:
This sounds fine.
Than you!

PR #1397, opened 2015-10-20
o.c.opibuilder: allow scripts to execute commands with parameters

willrogers:
Previously scripts could execute Eclipse commands.  Now they may
create ParameterizedCommands.


berryma4:
closes #1396

Issue #1399, opened 2015-10-20
PVTree fails if top node in alarm state


nickbattam:
PVTree (org.csstudio.diag.epics.pvmanager.pvtree) fails to discover the tree if the PV it is looking at is in an error state.

The top node is displayed as:
<pv_name> (ai [MAJOR,LINK_ALARM]) = 0 [MAJOR,LINK_ALARM]

The PV analysis assumes the record-type string just contains 'ai'.

kasemir:
Both that this error only applies to the `pvmanager` version.
The plain `org.csstudio.diag.epics.pvtree` is not affected.

PR #1400, opened 2015-10-20
Issue 1399 pvtree

nickbattam:
Issue #1399 - strip alarm state information from record type string before processing links


kasemir:
Thanks for fixing this.

Issue #1404, opened 2015-10-27
PVTree doesn't follow SCALCOUT record links


nickbattam:
The PVTree (org.csstudio.diag.epics.pvmanager.pvtree) does not trace links out of in SCALCOUT records.

PR #1405, opened 2015-10-27
#1404 Add scalcout record links

nickbattam:
Add missing scalcout record type to preferences file


berryma4:
@kasemir okay with this?

kasemir:
Yes, fine.

Issue #1406, opened 2015-10-28
Diirt Configuration


shroffk:
There are some issues regarding how to configure the new diirt datasources
- [x] Diirt configuration preference page
      - preference to define location of diirt config directory
      - support eclipse platform urls
- [x] Help documentation is needed on how to configure diirt
- [x] remove the references to the old pvmanager plugins and features and replace them with diirt plugins and features (https://github.com/ControlSystemStudio/org.csstudio.product/issues/14)

shroffk:
Created the Diirt preferences page which lets you define the location of the diirt configuration folder.

In addition it has a directory view which shows all the individual configuration files. Double click on any file will result in opening that file with the configured default editor (so you don't have to go searching for it).

![image](https://cloud.githubusercontent.com/assets/2111304/10790326/01ad0840-7d59-11e5-97ab-e67638e99d8b.png)

![image](https://cloud.githubusercontent.com/assets/2111304/10790342/1880491a-7d59-11e5-866a-7ae0da7327fa.png)

shroffk:
Note:
@willrogers has already worked on supporting the platform URLS http://help.eclipse.org/juno/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Frcp_perspective.htm

Do we want to support these or use the format requested by @kasemir 
````
org.csstudio.diirt.util.preferences/diirt.home=/etc/diirt
org.csstudio.diirt.util.preferences/diirt.home=${user.home}/.diirt
org.csstudio.diirt.util.preferences/diirt.home=${osgi.install.area}/diirt
org.csstudio.diirt.util.preferences/diirt.home=${osgi.configuration.area}/diirt
org.csstudio.diirt.util.preferences/diirt.home=${osgi.instance.area}/diirt
````

kasemir:
Not sure about the rcp_perspective link you mention, but I'm fine with platform URLs.

They would cover what I really need: install location.
They'd also cover instance or config location.

They wouldn't allow specifying the user's home directory, but when we look for "specific to user" I think instance location (= workspace) is basically the same idea.

So: Yes, platform URLs are OK.

kasemir:
I found this link on platform URLs: http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fplatform-scheme-uri.html

[...]

PR #1407, opened 2015-10-28
Diirt config

willrogers:
This should understand the `platform:` URIs and also not complain if it doesn't recognise the contents as a directory.

Example configuration: `platform:/base/configuration/diirt_5064`.

As discussed in #1406.


willrogers:
Kunal's solution is more advanced!

Issue #1401, opened 2015-10-20
PVTree displays array data as classname


nickbattam:
PVTree (org.csstudio.diag.epics.pvmanager.pvtree) does not display array data correctly. For PVs providing array data the tree shows the diirt class name rather than value e.g. org.diirt.support.ca.VIntArrayFromDbr.

The diirt source implements a .toString() method for each data type, including those which extend VNumberArray, but this is not being used. The rendering is being done by org.csstudio.diag.epics.pvtree.VTypeHelper which cannot handle array data.

nickbattam:
The primary piece of data my users want in the PV tree is the alarm state. 
They are happy the first n elements of the array to be displayed, where n can be 1.

nickbattam:
Closing resolved issues

PR #1409, opened 2015-10-28
Extend PVTree to write first-value and size for number arrays

nickbattam:
See #1401


shroffk:
No, The assumption is correct.
The code looks correct and I think it should be merged.

PR #1413, opened 2015-10-29
Issue 1406

shroffk:
@willrogers @kasemir @berryma4 @utzeln 

Issue #1406 
Adding a preference page for diirt configuration folder.

Note:
do not merge until behaviour is verified

### DIFF
--- a/core/diirt/diirt-plugins/org.csstudio.diirt.util.preferences/build.properties
+++ b/core/diirt/diirt-plugins/org.csstudio.diirt.util.preferences/build.properties
@@ -2,4 +2,5 @@ source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               plugin.xml
+               plugin.xml,\
+               preferences.ini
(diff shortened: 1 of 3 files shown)


kasemir:
I'm for merging this.

berryma4:
It's great for me.  @willrogers did you have anything else to add?

willrogers:
This is very good and makes the change to Diirt more manageable.

I'm getting some problems with the 'Apply' button, though.  Sometimes when I press it the message is 'null argument:'.  Occasionally I get a pop-up error.  Can anyone reproduce this?

Bonus feature suggestion: should the tree table show the title of the directory it is showing?  Sometimes I've changed the configuration and it's not clear to me whether the tree table is showing the new location or the old one.  That's picky though because this is a good feature, thanks Kunal.

shroffk:
@willrogers the tree table only updates when you 
1. select a new folder from the dialog
2. if you manually type the address and then hit the apply button (the return usually will execute the "Ok", which is in focus by default)

but...I agree we need a way to indicate what is being displayed.

The null argument is because of me expecting exceptions to give reasonable messages...of course I am wrong with this assumption :)...I guess I should not be lazy and put the message in myself.

I have added more explicit error messages and added a check to ensure that the pop-up error is fixed

willrogers:
Thanks, looks good.

Issue #1281, opened 2015-07-31
Alarm Table: alarm description without alarm severity prefix


utzeln:
Would it be possible to make the alarm severity prefix optional [by default: true]?

On ITER, as we intend to use the alarm icon and the alarm message background colour to reflect the alarm severity, we would like to reduce the Description column to what was configured by the user.

kasemir:
When the description text starts with "* ", the "MINOR Alarm: " prefix is omitted, so I think you can already get what you want.
See online help of alarm system 10.2 Description, which points to 11.3 Annunciation Text.

utzeln:
Right - this means that we ask our user to add * to the description of the alarm.

I thought that this was only for the annunciator.

kasemir:
Yes, description and annunciation are currently pretty much the same.

OK, so do I understand correctly that you'd like to hide the "MINOR Alarm:" from the description column in the table, since the severity is already obvious from the separate Severity column and/or icon and/or future background color, but the annunciation would remain unchanged?
Annunciation still defaults to "MINOR Alarm: whatever", unless the description starts with "* "?

I think that'll be good, we don't even need a preference.
Just change the table to always show just the description.

Technical detail: The description could be "* The water temperature reached {1}" where {1} will be replaced by the value that triggered the alarm.
I assume you'd like to show "The water temperature reached 110" and not "The water temperature reached {1}"?
So the change wouldn't just show the plain description input. It still needs to format it, but using a format that slightly differs from the one used for annunciation, omitting the prefix and the "* " handling.

utzeln:
Yes - the alarm annunciator and the alarm notifier will make use of the * prefix and by default always add the severity information, except told otherwise.

And right, the description can include {0} and {1} that have to be formatted - the alarm description is not just plain text.

Thanks.

Issue #1119, opened 2015-05-12
The background colour of the alarm message shall reflect the priority of the alarm


utzeln:
This should be an option set by default to false.

Now what happens for alarm that clears without the operator acknowledging it?

If an alarm clears before acknowledged by the operator, the alarm shall remain in the list (with a different alarm severity icon)  but will be presented in reverse-video (text font color becomes
background color, and background color used for the alarm font) - for instance red message on white background instead of white message on red background:
![image](https://cloud.githubusercontent.com/assets/4686181/7584713/a9f8061a-f89f-11e4-8c07-ecbb70f8fba8.png)

The current severity is not used for animation.

utzeln:
Updated alarm coding for ITER:

![image](https://cloud.githubusercontent.com/assets/4686181/8984284/e6e6ac7e-36d0-11e5-9ca0-ecbfaeaffaea.png)

PR #1422, opened 2015-11-03
Alarm table: description & severity sensitive background

jbobnar:
#1281 Implementation of the alarm description presentation without the alarm severity prefix
#1119 Implementation of the "severity sensitive background" - default preferences should give the same result as before.


berryma4:
@kasemir can you review

kasemir:
Fine.

PR #1427, opened 2015-11-04
Alarm table opi widget

jbobnar:
Two separate properties have been added to the alarm table opiwidget, which allow to hide/show the column headers (titles) and the table headers (number of alarms, filter dialog). This can only be configured in the opiwidget. The "standalone" alarmtable is non-configurable.

Let me know if I need to create an issue for this.


berryma4:
@kasemir can you test.
Thank you!

kasemir:
Can't test the table opiwidget since I'm not using that and unclear how it's supposed to act, but the changes to the original alarm table appear OK, so go ahead.

PR #1428, opened 2015-11-04
Extend Standalone window manager to include Show All button

nickbattam:
Running the CSS application window at full screen, each click on a 'Show' window button pops one of the windows to the top, but because the CSS window also comes to the top of the stack, any other windows are pushed back.

This adds a 'ShowAll' button which brings all managed windows to the top.


berryma4:
@shroffk can you test.
Thank you!

PR #1430, opened 2015-11-04
Issue 1424

nickbattam:
@shroffk This now handles 'null' paths in the tree display and have added crude validation of the user's folder selection.


shroffk:
I reviewed the changes and they can be merged.

However, some of these fixes should be handled by diirt...including create a datasource.xml if it is missing and using the default configuration files that are included as resources in the jars themselves.

nickbattam:
@shroffk I think my 'verifyDiirtFolder' method could/should be moved into the Diirt source. This would remove the 'knowledge' of datasources.xml from the properties panel. 

As far as I can see from the existing code
> " some of these fixes should be handled by diirt... including create a datasource.xml if it is missing
> and using the default configuration files that are included as resources in the jars themselves."

is already handled by DataSourceProvider in diirt

PR #1461, opened 2015-11-23
Issue 1412 bytewidget

nickbattam:
Extended LED widget to have configurable 'borders', in addition to the widget's regular draw2d 'Border', which is square for round LEDs. This replaces a fixed 3px grey border.

The byte widget exposes the LED 'border' width and color allowing it to be configured to display bits at a higher density.


berryma4:
@shroffk

shroffk:
This passes the existing LED regression tests.

Based on the code I am guessing the new properties that need to be tested are bulb border and bulb border color.
Both properties behave as expected.
image

In the future it would be nice to have the bulb border function as the primary and only border which is also alarm sensitive.

Since I can't find the associated Issue ticket I am not sure if I have tested everything, please let me know if I missed something.

berryma4:
@shroffk thank you!

Issue #1467, opened 2015-11-25
Bytewidget layout changes


nickbattam:
Changes made to the ByteWidget has altered the default rendering and LED size and close-packing of LEDs does not render well in 3D.

I will add a 'pack' property to the widget which defaults to 'False'. This will restore the original rendering behavior.

Issue #1412, opened 2015-10-29
Byte monitor widget does not scale well


nickbattam:
I have a number of screens which use ByteMonitor widgets (org.csstudio.swt.widgets) to display 8+ bits of data in a relatively narrow column (e.g. 11 bits of data in 51 pixels). At this density the green part of the LED is of zero-width.
The equivalent EDM control is able to display data at this density, using a single pixel border between cells.

Looking at the implementation, the byte monitor widgets is constructed from a number of LEDFigure widgets (org.csstudio.swt.widgets) placed in a vertical or horizontal line. Each (square) LEDFigure in the byte monitor widget is drawn with an unconfigurable, hard-coded, 3px border. This gives a 6 px gap between the green part of adjacent cells.

Two fixes are required:
- update the LEDFigure to respect the 'border' property
- update the Byte Widget to set smaller borders between adjacent cells

I propose to add a new 'spacing between bits' property the byte widget and use that to configure the correct border on the LEDFigure rather than attempt to determine an appropriate space algorithmically.
 Changing the layout of the LEDFigures in the byte widget so borders of adjacent LEDs are overlapping would allow a single pixel spacing.
If the default property value was left at 6px this should be backwards compatible with existing OPI panels.

PR #1468, opened 2015-11-25
#1467 Add 'pack' LEDs option to ByteWidget

nickbattam:
Revert behavior change introduced by fix to #1412.

Added new property that reproduces previous layout/spacing of LEDs inside a byte widget. Defaults to rendering with legacy layout.


berryma4:
#1467

Issue #1073, opened 2015-04-22
BOY: Add icon to Menu Button


mfurseman:
When controlling an Enum PV in EDM the menu button has a rectangular icon hinting to users that it will open a drop down list:
![sofb-edm](https://cloud.githubusercontent.com/assets/5279818/7273231/7e6bc790-e8e9-11e4-8c22-0659c566cd45.png)
We would like this functionality in CS-Studio, however the menu button does not provide the option to choose an icon. After looking at implementing this it seems that there would be few differences remaining between the MenuButtonFigure and ActionButtonFigure. 

Although it is relatively easy to just add an icon to the menu button figure, does it seem like unifying these figure classes makes sense?

kasemir:
If you can merge the two button figures, I'd say go for it.

As for the specific use with enums and drop-downs, I agree that a visual hint is good.
Do you want to automate this, i.e. menu buttons would always have that icon?

I wouldn't want to limit us to the Motif 'square'. Even past EDM users should be OK with a more common 'drop down' hint, for example the one used by SWT:
![image](https://cloud.githubusercontent.com/assets/1932421/7277342/7c00b990-e8dd-11e4-8c5b-a4302bd277ec.png)  Or goggle for 'drop down menu' images.

mfurseman:
I'm not a fan of the Motif square ethier!

We could either choose a graphic to bundle and display on the button (perhaps with an option in the model to hide/show it), or just let the user choose an icon to include as the Action Button does. In the interest of presenting the user with a consistent interface I would prefer to always show a fixed icon, but having created this issue ticket I am sure to be biased.

berryma4:
:+1: for the google image (a copyright safe version at least)

kasemir:
I'd vote for 'fixed' icon, too. As for the specific icon, I'd either use

a) What SWT uses on the platform for the SWT.ARROW button, Combo drop-down etc. See
https://www.eclipse.org/swt/widgets/ That makes it consistent with the overall look of everything else on that platform. .. but it will result in a different icon for Windows, Linux, Mac OS X. Plus for people who really want everything to look like EDM: If you resurrect the build of SWT for Motif, it'll look 100% like EDM.

b) Some generic 'down pointing triangle' icon like 
[...]

PR #1474, opened 2015-11-26
Issue 1073: Add icon to Menu Button

mfurseman:
This (finally) provides the fixes requested in issue #1073. The icon provided is a generic down arrow, with an additional option to show the arrow in the properties. The new option defaults to false so that the behaviour of existing displays is retained.


berryma4:
Thank you!

Issue #1470, opened 2015-11-25
remove old pvmanager references


berryma4:

berryma4:
relates #1171

PR #1475, opened 2015-11-27
o.c.graphene.opiwidgets: update reference to org.epics

willrogers:
This reference to `org.epics` was breaking the build.


berryma4:
@shroffk might care about pretune.  I believe the utilities feature I remove two days ago was the only feature that included graphene.opiwidgets.  We should probably make a new feature for it.

berryma4:
#1470

shroffk:
I will create the new features today

berryma4:
Done, #1476

Issue #1495, opened 2015-12-04
.opi files have mixed line endings


tboegi:
It seams as if CSS stores opi files with "mixed line endings":
The first line is terminated with a line feed (LF or "\n"):
<?xml version="1.0" encoding="UTF-8"?>

The next lines are terminated with carriage return line feed (CRLF, "\r\n"):
<display typeId="org.csstudio.opibuilder.Display" version="1.0.0">^M

And the last line has LF.

Is there a reason to have CRLF in some lines ?

willrogers:
This seems to be true and is indeed very strange.  There is no reason for this behaviour.

I think we should write Unix-style line endings.

berryma4:
I agree!  Can we make the elements alphabetized while we are at it? (So
diffs on opis aren't as crazy as they are)

PR #1496, opened 2015-12-04
o.c.opibuilder: use Unix-style line endings in OPI files.

willrogers:
As discussed in #1495.

This should be reviewed by at least one other @berryma4 @shroffk @kasemir

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/persistence/XMLUtil.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/persistence/XMLUtil.java
@@ -15,6 +15,7 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


kasemir:
Seems to handle the line endings.
As Eric suggested, while at it, do you see an easy way to get reproducible order of widget properties? Sort them alphabetically instead of quasi-random order of map keys?

willrogers:
It will need a bit more care.  This is where the properties are seirialised from within widgets.  It should be easy enough to create a sorted list of properties here.

I assume that the order of widgets within opi files is well-defined already.  Is that true?

    public static Element widgetToXMLElement(AbstractWidgetModel widgetModel){

        Element result = new Element((widgetModel instanceof DisplayModel) ? XMLTAG_DISPLAY :
            (widgetModel instanceof ConnectionModel) ? XMLTAG_CONNECTION : XMLTAG_WIDGET);
        result.setAttribute(XMLATTR_TYPEID, widgetModel.getTypeID());
        result.setAttribute(XMLATTR_VERSION, widgetModel.getVersion().toString());
        for(String propId : widgetModel.getAllPropertyIDs()){
            if(widgetModel.getProperty(propId).isSavable()){
                Element propElement = new Element(propId);
                widgetModel.getProperty(propId).writeToXML(propElement);
                result.addContent(propElement);
            }
        }

willrogers:
I pushed a commit, but I haven't tested it!

kasemir:
Yes, that sorting of properties should help.

Yes, the order of widgets is defined and must not be changed.The order matters in terms of 'in front of' or 'behind'. In principle, it also defines the'tab order' when you use the keyboard to navigate between widgets. That only works with native SWT widgets, but still, changing the widget order would go the wrong direction.

kasemir:
Regarding your test, not sure how to really test the case where the order of map key entries changed. The order of properties tended to stay the same for a simple rectangle move on the same computer with the same version of CSS anyway. The order tended to change as you edited the same file on a different OS, or with a different build of CSS. But besides the test, a plain code comparison shows that we go from undefined map key order to Collections.sort(make keys), so I'm quite sure it's fine.

Issue #1489, opened 2015-12-01
opi-regression-tests - graphene examples


berryma4:
Tables, bubble charts, etc. are not displaying
![image](https://cloud.githubusercontent.com/assets/854021/11509809/59a9237c-982e-11e5-8e8d-9b7c5f9b4df3.png)

berryma4:
I've added it to the general product:
https://github.com/ControlSystemStudio/org.csstudio.product/commit/8d722bf0f1824e9b0bfa7bcd3b5340c2ab024f83

PR #1498, opened 2015-12-04
#1489 creating a new feature for graphene opiwidgets

shroffk:
A feature for including graphene widgets into cs-studio
This feature is need in order to test the graphene widgets in opi-regression-tests


berryma4:
#1489

Issue #1510, opened 2015-12-10
Update versions in opibuilder plugins


willrogers:
We need to update version in opibuilder plugins.

I can do this but I'm guessing about plug-ins I don't know about.  Typically I've given a major version bump to any plugins that have changed since 4.1 and not changed any plugins that haven't.

I also haven't touched rap plugins because I don't know what's going on there.

PR #1511, opened 2015-12-10
Update versions on opibuilder plugins and features.

willrogers:
Someone needed to do this so I've given it a go.

I have updated opibuilder to 5.0.0.  I don't mind that it's not the same as the CSS version and it is definitely a major update.

I haven't changed rap plugins because I don't understand them.

If a plugin has changed I've updated the major version.  If a plugin hasn't changed I **haven't** updated the version.

I don't mind if people want to do this differently @kasemir @berryma4 @shroffk .

See #1510.

### DIFF
--- a/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
+++ b/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.csstudio.applications.opibuilder.feature"
       label="Operator Interface Builder (BOY)"
-      version="4.0.107.qualifier"
+      version="5.0.0.qualifier"
       provider-name="Kay Kasemir &lt;kasemirk@ornl.gov&gt; - SNS">
 
    <description>
--- a/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/pom.xml
+++ b/applications/opibuilder/opibuilder-features/org.csstudio.applications.opibuilder.feature/pom.xml
@@ -5,9 +5,9 @@
   <parent>
     <groupId>org.csstudio</groupId>
     <artifactId>opibuilder-features</artifactId>
-    <version>0.0.1-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
   </parent>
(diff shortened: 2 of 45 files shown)


berryma4:
thank you!

shroffk:
Thanks for updating all the version numbers Will.

In the future do you think it makes sense to manage the version numbers at an entire module level (logbook, alarm, opibuilder) using maven release?

We could then use maven to manage the version numbers, such that with each release build of that module all the sub modules get automatically updated.
http://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html

but we have to agree that all sub modules will have the same version numbers.

willrogers:
Kunal's suggestion would make keeping the version numbers up to date so much easier that I am inclined to agree.

I can't think of a situation where it would be important to have separate versions for different plugins in a module.

kasemir:
Does the maven release mechanism overwrite the version info inside the pom files? Or does this require using variables?

If the maven release mechanism overwrites the underlying pom file versions, I'm all for it.
That way we can usually release the whole module as version 4.1.0, 4.2.0, etc.
But if for some reason one plugin needs to be updated in between, we could manually update just plugin to 4.1.1, 4.1.2, .. until the next whole module update that goes to 4.2.0.

berryma4:
We are using the maven/tycho update versions in org.csstudio.product.  It works nicely, modifying the poms and manifests.  I not sure how to use the maven release plugin, because it doesn't update the manifest.  Maybe we could tie the tycho update versions to a phase in maven release?  Is that possible?

We have discussed versions before, but we should make this an official project for 4.3.x, and call this version change as okay on 4.2.x.

shroffk:
I guess this is the plugin we should use is
https://eclipse.org/tycho/sitedocs/tycho-release/tycho-versions-plugin/plugin-info.html

This plugin should manage the version numbers in both the pom.xml and the manifest files.
I think the tycho version plugin will work as you described Kay, the versions will be updated only once we do a release, and individual plugins and feature versions can be managed manually if need be.

berryma4:
I use this for the org.csstudio.product, examples:
https://github.com/ControlSystemStudio/org.csstudio.product/blob/master/prepare-release.sh
https://github.com/ControlSystemStudio/org.csstudio.product/blob/master/prepare-next-release.sh
(although it doesn't move .product versions up)

Issue #1516, opened 2015-12-11
Save Restore


berryma4:
![ezgif com-video-to-gif](https://cloud.githubusercontent.com/assets/854021/11745539/e11c4338-9fe6-11e5-8947-d20f80cabfef.gif)

Code at:
https://github.com/frib-high-level-controls/save-set-restore

supporting documents on cs-studio google drive / projects:
https://drive.google.com/drive/folders/0B_ikNAN_jxjPM001TmNMd25pNk0

Currently synchronizes with git repo (local and remote).

So changes in the next iteration:
 - extension point to add compare of other related values (ie. optics, read value)
 - allow adding pvs to save-set, checking archiver for values
 - savable filters for restoring save set
 - sort / maybe search tags
 - right click context menu to send pvs to databrowser at that timestamp
 - send to logbook
 - periodic chart feature is not general, so a general text string list feature will be added
 - masar data provider
 - scale pv value before applying sets (API)
 - read only pvs can be added for comparison
 - pvs context from other applications can be used to load configuration set of pvs

berryma4:
The diirt archive datasource; What object does it return?

berryma4:
merging to applications

berryma4:
#1789

berryma4:
project has been merged to applications, any problems can be new issues.

PR #1534, opened 2015-12-21
Archive diirt

jbobnar:
Implementation of the archive DIIRT datasource. 

You can now do:
<b>archive://\<archiver\>/pvName</b>, where <i>\<archiver\>/</i> is an optional parameter providing the url or name of the archiver as specified in the archive.xml diirt configuration. If not provided all archivers are used. This will create a channel, and you can write the time parameters to tell the channel what it should fetch. For example:
```
PVWriter writer = ...
writer.write(Boolean.FALSE); // set raw data retrieval
writer.write(new Date[]{startTime,endTime}); //fetch data
// fetch data using optimised data retrieval
writer.write(new Object[]{startTime, endTime, Boolean.TRUE});
//startTime and endTime can be given as long, Date, Timestamp, Time; 
//see DynamicArchiveChannelHandler#write for more info
```
The channel is multiplexed, therefore if you have many PVReaders attached to the same pv and you do write on any of them, it all boils down to a single data provider.

Another way to create a channel is also:
<b>archive://\<archive\>/pvName[startTime;stopTime;optimised)</b>. This will create the channel and immediately fetch the data. Once created you cannot change any parameter. This channel is also multiplexed, however, the parameters in brackets are considered as part of the unique channel name.
  
Configuration of the diirt datasource is done via xml. There is an example xml located in the folder rootfiles_example. You need to place this file into the diirt/datasources/archive folder.


berryma4:
could
```
archive://<archive>/pvName[startTime;stopTime;optimised)
```
be uri like:
```
archive://<archive>/pvName?startTime=123123&stopTime=123123&optimised=true
```

jbobnar:
Updated the parameters syntax. You can now provide them like you suggested. 

I also updated the way how the <archive> is treated. If it is provided as the name of the archive source defined in the xml, that source will be used:
```
archive://myArchiveSource/pvName?startTime=123123&stopTime=...
```
If no such source is defined, all configured sources are used.
If \<archive\> is provided as a url (it contains "://") that url will be used regardless if is defined in the diirt configuration or not. For example:
```
archive://pbraw://myArchiveServer:17668/retrieval/pvName?startTime=123123&stopTime=...
```

berryma4:
``` archive://pbraw://myArchiveServer:17668/retrieval/pvName?startTime=123123&stopTime=...```

scheme://scheme://host seems odd.  Just pbraw:// seems more natural.  But, this doesn't fit the archive plugin.  Maybe make it another query parameter, protocol=pbraw, if archive stays the scheme.  Or, url=  pbraw://myArchiveServer:17668/retrieval/pvName?startTime=123123&stopTime=....

I feel the options are; pbraw://, or archive://direct?url=pbraw://myArchiveServer:17668/retrieval/pvName?startTime=123123&stopTime=...., or not supporting this direct calling at all, if using archive as the scheme.  @shroffk, your thoughts?

... I really wish we were using the diirt cache://, but this is a nice first step.

**Comment by jbobnar** [comment]
I wanted to have a way to use an archive source that is not defined in the diirt configuration. I know scheme://scheme:// is not the a common thing, but it is easy to parse: anything after last '/' is the pv name.

pbraw://something is impossible, because then we would need a separate datasource for every archiver type (diirt takes away the leading schema definition).. One also loses the independence from the archiver implementation - the client would need to know the type of the archive to use. If you ever switch to another archiver, you'd need to update all occurrences of pbraw:// to something else.

The second option seems rather complex, having two levels of parameters... In such case I'd rather have:
```
archive://pvName?startTime=123&stopTime=223&url="pbraw://myarchive:17668/retrieval"
```

**Comment by berryma4** [comment]
I think this is the strange part: "One also loses the independence from the archiver implementation", yet wanting to add a direct url.
```
archive://pvName?startTime=123&stopTime=223&url="pbraw://myarchive:17668/retrieval"
```
I meant keeping the word direct, literally:
```
archive://direct&url="pbraw://myarchive:17668/retrieval/pvname"
```

Ah, maybe I'm confused, ... do you want to add a supported archive source, or make a direct call to an archive?

I think we should probably not support either.

**Comment by jbobnar** [comment]
I just wanted to add a source that is not defined. There should never be any direct calls. That could get weird. 
My thinking was: you have the list of data sources defined in the preferences (current situation). One can offer the user a list to pick the source and fetch the data for that source without having the diirt configuration in order. Having that one could easily sneak this into the databrowser.

Currently this is implemented, but I can always comment it out.

**Comment by berryma4** [comment]
so it would be:
```
archive://pvName?startTime=123&stopTime=223&url[]="pbraw://myarchive:17668/retrieval"
```
with url being an array.  That sounds fine.

**Comment by jbobnar** [comment]
Done.
```
archive://pvName&startTime=123...&url[]="pbraw://..."&url[]="pbraw://..."
```
The quotes around urls are optional. They are only needed if the archive url has its own parameters as well. This is not the case for any existing archiver, but maybe there will be something like that in the future.

berryma4:
relates to #1516

berryma4:
could we add a subtree of this to https://github.com/frib-high-level-controls/save-set-restore ?
or subtree to its own repo at https://github.com/frib-high-level-controls/
So we can use/test it on 4.2

PR #1541, opened 2015-12-24
Fixed the name of the feature project and added some value type checking

jbobnar:
Fixed a copy-paste bug.
Added some value checking.


berryma4:
Merry Christmas!

Issue #1520, opened 2015-12-15
Diirt configuration and help documentation


shroffk:
Provide documentation on how the diirt configuration works with respect to cs-studio.

shroffk:
When cs-studio product is started the first diirt module to start up is the org.csstudio.diirt.util (This is the module that provides the description for the osgi services associated with datasources, functions, and formulas)

```
Dec 15, 2015 8:57:29 AM org.csstudio.diirt.util.Activator start
INFO: Starting diirt.util
```
Since the preferences are not yet available, diirt uses the config location defined in the jvm system properties, if you want a different default from the one described you can manually set the jvm property "diirt.home" in your products launch configuration

```
Dec 15, 2015 8:57:29 AM org.diirt.util.config.Configuration getFile
INFO: Loading C:\Users\Kunal Shroff\.diirt\datasources\datasources.xml
```

Following this all the Datasources Providers - provided via java services are made registered
```
Found bundle org.diirt.datasource_3.0.1.20151009125656 [880]
Found bundle org.eclipse.osgi.internal.framework.BundleContextImpl@43bf3668
Found bundles [Lorg.osgi.framework.Bundle;@4c6f307e
Found bundleentry://809.fwk741191984/META-INF/services/org.diirt.datasource.DataSourceProvider in org.diirt.datasource-loc_3.0.1.20151106105554 [809]
Providers [org.diirt.datasource.loc.LocalDataSourceProvider]
Provider instance org.diirt.datasource.loc.LocalDataSourceProvider@362ba6cc
Found bundleentry://884.fwk741191984/META-INF/services/org.diirt.datasource.DataSourceProvider in org.diirt.datasource-sim_3.0.1.20151009125622 [884]
Providers [org.diirt.datasource.sim.SimulationDataSourceProvider]
Provider instance org.diirt.datasource.sim.SimulationDataSourceProvider@42427546
Found bundleentry://885.fwk741191984/META-INF/services/org.diirt.datasource.DataSourceProvider in org.diirt.datasource-sys_3.0.1.20151009125624 [885]
Providers [org.diirt.datasource.sys.SystemDataSourceProvider]
Provider instance org.diirt.datasource.sys.SystemDataSourceProvider@69c2cc9c
Found bundleentry://889.fwk741191984/META-INF/services/org.diirt.datasource.DataSourceProvider in org.diirt.support.diirt-ca_3.0.1.20151009125648 [889]
[...]

PR #1552, opened 2016-01-06
#1520 the configuration directory set during preWindowOpen stage

shroffk:
Temporary solution to the diirt configuration issue.

Could you review
@berryma4 @kasemir @willrogers @jbobnar 

Note:
This should re enable the workspace prompt and allow the config folder to be set via the diirt config folder preference
The xml defined services have not been tested

### DIFF
--- a/core/diirt/diirt-plugins/org.csstudio.diirt.util/META-INF/MANIFEST.MF
+++ b/core/diirt/diirt-plugins/org.csstudio.diirt.util/META-INF/MANIFEST.MF
@@ -9,7 +9,6 @@ Service-Component: OSGI-INF/ds_consumer.xml,OSGI-INF/ffs_consumer.xml,
 Import-Package: org.diirt.datasource;version="3.0.1",
  org.diirt.datasource.formula;version="3.0.1",
  org.diirt.service;version="3.0.1",
- org.diirt.util.config;version="3.0.1",
  org.eclipse.core.runtime;version="3.4.0",
  org.eclipse.ui,
  org.osgi.framework;version="1.8.0"
@@ -18,5 +17,6 @@ Bundle-Activator: org.csstudio.diirt.util.Activator
 Require-Bundle: org.eclipse.equinox.preferences;bundle-version="3.5.200",
  org.eclipse.osgi,
  org.csstudio.startup;bundle-version="3.2.1",
- org.csstudio.utility.product
+ org.csstudio.utility.product;bundle-version="1.3.0",
+ org.diirt.util;bundle-version="3.0.1"
 Export-Package: org.csstudio.diirt.util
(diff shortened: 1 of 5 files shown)


kasemir:
Do you know what actually accessed the diirt code and thus caused the diirt Activator to run before the workspace was selected?

In any case, moving the init code from the activator where we're not quite sure when it's invoked to the IWorkbenchWindowAdvisorExtPoint looks good to me.

berryma4:
can we get this on 4.2.x?

jbobnar:
I think this will work, but can you update the diirt p2 repository with the new org.diirt.util plugin. It is mandatory to compile this code.

shroffk:
@jbobnar I am not sure I completely follow? I have merged the associated diirt code into the diirt master branch.

@kasemir I think it was the fact that this plugin also contributed the osgi service descriptions, since the  org.eclipse.equinox.ds start up level is 2 (this is recommended by eclipse)...this in turn searches for and registers all osgi services which results in the activation of the diirt plugin

kasemir:
OK, the service registration sounds plausible. So whatever happens at that time must avoid access to preferences.

berryma4:
4.2.x please?

jbobnar:
@shroffk 
If I execute this:

eclipsec -application org.eclipse.equinox.p2.director -repository http://diirt.download.controlsystemstudio.org/diirt/3.0.1 -list "Q:everything.select(x | x.id == 'org.diirt.util')"

I get the following answer: org.diirt.util=3.0.1.20151009125657
Which means that the p2 repository still has the diirt plugin from 9th October 2015. It should be updated.

berryma4:
Yes, ... @shroffk is having problems with Cloudbees, where the release is done.  The new release will be 3.0.2.  I should have waited to merge this until that was done (with the repo url fix).

I have the 4.2.x branch pull request, cherry picked, coming this morning.

shroffk:
@jbobnar I opened an issue with Cloudbees yesterday...they have escalated the issue but there has been no resolution.
I will wait till afternoon otherwise manually try to update p2 repos.

kasemir:
Thanks for re-enabling the workspace selection, I really needed that!

For what it's worth, I have a cron jobs that install the 'nightly' update of CSS by downloading the basic SNS CSS and then invoking it with `-application org.eclipse.equinox.p2.director` to add optional features. The log shows that the diirt Activator starts even for those:

```
Jan 06, 2016 3:10:50 AM org.csstudio.diirt.util.Activator start
INFO: Starting diirt.util
Installing org.csstudio.sns.feature.ccr.feature.group 4.0.0.201601052111.
```

Not a problem, but shows that there's little control over the service lookup schedule.

Issue #1429, opened 2015-11-04
PV Layer roadmap


kasemir:
Illustrations that aim to summarize discussions amongst attendees of the Melbourne EPICS/ICALEPCS.

We currently have two PV layers:

![screen shot 2015-11-04 at 10 07 45 am](https://cloud.githubusercontent.com/assets/1932421/10941998/fae01e02-82db-11e5-99d8-fe0b50d1d4af.png)

`vtype.pv` performs direct notifications for received events, and it support get/put callback.
`PVManager` decouples the events, notifying on different threads. It cannot support get/put callback, but instead offers formulas and non-event based data sources like files.
Both eventually provide `Type` data, and there are examples for applications that need each one. 

The current parallel implementations result in different local PVs, possibly duplication of network connections, and long term code management effort.

Suggestion:

![screen shot 2015-11-04 at 10 07 52 am](https://cloud.githubusercontent.com/assets/1932421/10942001/023907f4-82dc-11e5-9b14-5ab5f6ede0bb.png)

Use only the `pvAccess` library for EPICS control system data, since it includes Channel Access.
Some applications use `vtype.pv` directly, others use the `PVManager` where an updated `vtype.pv` becomes the real-time data source for local, EPICS and maybe also many simulated PVs.
`PVManager` adds decoupling, formulas, and supports non-real-time data sources.

Possible implementation steps
- [ ] Update local vtype.pv to  support same features as PVManager's local data source
- [ ] Update simulated vtype.pv to support same range as PVManager's simulated data source
- [ ] Update vtype.pv's pva to the latest pvAccess release
- [ ] Extend vtype.pv or provide adapter that allows usage as PVManager data source
- [ ] In PVManager, use vtype.pv as data source and remove original ca, pva, loc, sim data sources

berryma4:
![image](https://cloud.githubusercontent.com/assets/854021/19086945/30ceb266-8a3e-11e6-9406-51b6b63c3839.png)

![image](https://cloud.githubusercontent.com/assets/854021/19086958/3955bcb8-8a3e-11e6-9987-fe99debb3b3d.png)

PR #1553, opened 2016-01-06
Pv 1429

kasemir:
More for #1429: noise waveform, 'epics://' prefix for really old displays


MichaelRitzert:
Is there any change to get at least ae55fb4 into 4.2.x?
as it stands, we have 4.1 and 4.3 support epics://, but not 4.2.x.

PR #1559, opened 2016-01-11
Updated the ITER copyright header from 2015 to 2016.

jbobnar:
Updated the ITER copyright header to include the year 2016.


kasemir:
I think we should merge this.

PR #1560, opened 2016-01-14
Support conversion even if the selection doesn't match the target

jbobnar:
This fixes a problem if a selection is of type that is an instance of the target type but is not exactly the same class.

For example: a selection might be an array of TimestampedPVs (which extends ProcessVariable). The client asks to convert the selection to a ProcessVariable[] array. This should be possible, because the selection is of the requested type. Previously the method returned an empty array. Now it tries to convert. There are of course limitations (details are in the comments).


kasemir:
OK.
Feel free to add your name as an @author, since you're now also responsible for this hack.

jbobnar:
@shroffk 
ReflectionUtil cannot be fixed like you suggest, because ReflectionUtil cannot "crawl". That would require the ReflectionUtil to instantiate the target class (from the string parameter). This can only be achieved if the target class is on the immediate classpath of that plugin. In all practical usecases that is never true (in OSGI environments). 

Why do you think there are unpredictable outcomes?

shroffk:
Sorry, wrong word choice.
I think what I should have said was the ReflectUtil usage needs to be addressed. 

Ok, Let me read up a bit...

berryma4:
So, I'm not sure about this.  The whole setup seems odd to me (ReflectUtil) ... but if we were to work with what is here ..... Taking the first element to convert seems like a bad idea.  But, it raises the question, what are  the cases we want to cover, and what is the desired outcome for those cases?  I feel this hasn't been answered.

berryma4:
Please merge master to this branch.  A Travis ci yml file was added for testing.

berryma4:
@jbobnar @shroffk I'm going to merge this.  I don't like it, but the problem isn't the pull request it's the class itself (the arrays).

shroffk:
Ok as I said before you can go ahead with the merge,
I feel that this "util" really does not buy us much.

Issue #1568, opened 2016-01-15
OPI Editor perspective should open if you open an OPI file in OPI editor


willrogers:
First-time users of CS-Studio typically go to File | New | OPI File and create an OPI.

An OPI editor opens, but the properties view isn't there.  It kind of looks like you can create an OPI, but you really can't.

This applies to me too: I often open an OPI file in the editor but then need to manually select the OPI Editor perspective.

I think OPI editor perspective should be chosen automatically if you open an OPI file in the editor.

kasemir:
I also often need to select the editor perspective manually as soon as I edit a display file.
But previous attempts to automatically switch perspectives for other cases (data browser, opi view) turned out to cause more harm than good.
The only case where I encounter the IDE switching perspectives on me is for debugging, where it's a "Do you want to switch.." dialog.

So maybe do that: Prompt for a switch, which still allows a user to decline if that's not what you want.

willrogers:
I think you're right, it's got to be one of those options (maybe 'Always do this?' or 'Never do this') attached to a preference somewhere.

Debugging in Eclipse is the classic example.

willrogers:
I can do this easily in the `init()` method of the editor, but:

* this isn't really the right place
* it prompts also on startup, which is unfriendly

Do you know of a better way to intercept the opening of a particular editor?

kasemir:
Well, trying to guess which perspective the user needs is difficult..

Still, this is for the editor, so people who simply execute existing displays won't be affected, only those who really need to edit displays. I can't think of a place other than the editor's init(). Maybe do this:
```
 .. init(..)
{
  IPerspectiveDescriptor persp =
        PlatformUI.getWorkbench().getActiveWorkbenchWindow()
                         .getActivePage().getPerspective();
  if (! persp.getId().equals("..opi editor perspective ID"))
  {
      .. dialog "Do you want to switch to the OPI editor perspective" ..
```

That way, those who do want to edit will get the prompt.
On startup, if you had the editor open on shutdown, that was very likely in the OPI editor perspective, so you won't be bothered with another dialog.

willrogers:
I asked the Dawn people for ideas, and there is a way to do it without having it in the `init()` method.  I think it's less invasive this way, and the preference could be set by a site to default to "Never", meaning that no-one sees the behaviour by default.

willrogers:
This is now done - feel free to close.

PR #1578, opened 2016-01-26
Suggest changing perspective on opening OPIEditor.

willrogers:
Don't merge this yet.

Tie to a preference so that choice can be remembered.  Set up listeners
on workbench components to ensure this is active for each new part.

See #1568.

Method suggested by Dawn devs: set up PartListeners that trigger on opening a new part.  In practice, this requires multiple listeners in order to ensure that the hook is active in all parts of the workbench.

Review @gerring @kasemir .

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.editor/META-INF/MANIFEST.MF
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.editor/META-INF/MANIFEST.MF
@@ -20,7 +20,8 @@ Require-Bundle: org.eclipse.core.filesystem;bundle-version="1.2.0",
  org.eclipse.help;bundle-version="3.6.0",
  org.jdom,
  org.eclipse.ui.workbench;bundle-version="3.106.0",
- org.eclipse.ui.views;bundle-version="3.7.0"
+ org.eclipse.ui.views;bundle-version="3.7.0",
+ org.eclipse.ui
 Bundle-Vendor: Xihui Chen<chenx1@ornl.gov> - SNS
 Bundle-ActivationPolicy: lazy
 Bundle-Activator: org.csstudio.opibuilder.editor.Activator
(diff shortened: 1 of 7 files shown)


kasemir:
How is that functionally different from adding similar checkEditorPerspective() code to the OPIEditor's init() routine? In either case, as an OPI Editor is opened, it checks the perspective and prompts for a change.  Adding it directly to the OPIEditor's init() is less code than adding a listener to every part in every window.

willrogers:
There's one difference:  this way *doesn't* trigger on CSS startup (not sure why, other than the `earlyStartup()` method says it is called after workbench initialisation).

There are cases that OPI editors will be open on shutdown - in particular, if you change perspective editors remain open in the editor stack.  Then when you restart you'll get prompted to change perspective, which might confuse people.

Ideally I'd like this to happen only when someone has actively chosen to edit an OPI, because then it should make sense.  That is, as long as the user understands perspectives, which I think is essential when using eclipse.  And it also requires a better message than I have so far.

kasemir:
Got it. The doesn't-trigger-on-CSS-startup behavior is good.

The only-when-someone-has-actively-chosen-to-edit-an-OPI question is hard.
For the debugger, somebody invokes a "Run in debugger" action.
The equivalent for us would be the navigator action that opens an editor, but I'm not sure if you can hook into that.

willrogers:
I think this commit does achieve that.  There are three cases I can think of:

* File | New | OPI ...
* Right-click in navigator | Open With | OPI Editor
* Right-click on runtime OPI | Edit OPI

These are all caught and are all deliberate actions.  I can't think of another way to open OPIEditor other than CSS startup, and we're skipping that now.

kasemir:
Right.

I'm OK with merging this.

There's no preference to disable it, but worst case you get the dialog once, check "don't ask again" and you're fine. Plus this only applies to people who edit, not beam line users who only execute existing displays.

willrogers:
Hold on before merging, I'll improve the messages.

@berryma4 @shroffk any comments on this feature?

I think a site could set the default preference to "Never", then no users at that site would see the dialog unless they changed the setting themselves.

willrogers:
The latest commit moves all state into the constructor.  This could be turned into an abstract class for any plugin that wanted to use this behaviour to extend.

willrogers:
We've built and tested this locally, and I'm happy for it to be merged @berryma4 .

berryma4:
Thank you!

Issue #1588, opened 2016-02-02
ProcessVariable -> DataBrowser Context menu does not work


jbobnar:
When clicking on a widget (e.g. text input widget) and selecting DataBrowser from the popup menu, the databrowser does not open, or if it does the data are not fetched for the selected PV.

willrogers:
Can we get this fix on `4.2.x`?

berryma4:
I'll cherry pick to 4.2.x

jbobnar:
This has been merged and cherry-picked to 4.2.x as well.

kasemir:
Thanks for identifying and fixing this!

Since the context menu had been OK for many years, any idea how the problem got introduced?
Is the issue only about the 'ChannelInfo' internal to the data browser, or do other handlers of PVName context menus also need to be updated?

jbobnar:
@kasemir 
It got broken by this commit dc7fd26b - getActiveMenuSelection was replaced by getCurrentSelection. I guess getCurrentSelection works OK for most structured viewers.

willrogers:
Yes, it must have been introduced by my commit, so sorry for that.  I think I tested only from the databrowser view, where the current selection was the same as the active menu selection.  Jaka's fix will also avoid `ClassCastException` in the apparently unlikely event that what is retrieved is an `ISelection` but not an `IStructuredSelection`.

PR #1589, opened 2016-02-02
Fix for databrowser context menu issue #1588

jbobnar:
This is fix of the issue #1588.


applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/OpenDataBrowserPopup.java
```diff
@@ -44,8 +45,11 @@
     public Object execute(final ExecutionEvent event) throws ExecutionException
     {
         // Get selection first because the ApplicationContext might change.
-        // This works for both context menu and double-clicks.
```
berryma4: Does this mean it no longer works for double clicks?

PR #1595, opened 2016-02-05
Rap3.0 & Diirt

jbobnar:
I updated all the RAP stuff to RAP3.0 and replaced pvmanager references with diirt (in rap features). 

Please, do not get scared of the number of files. The majority of changes in the common plugins are the result of the fact that eclipse guys diverged the RAP API from the RCP API. This broke the "single-sourcing design" for a few classes: ColorConstants, SchemeBorder and ButtonBorder. There are no functional changes to the common code.

The RAP plugins have been updated to work with RAP 3.0 and the rap.rtplot has been synchronised with the swt.rtplot.

Issue #1594, opened 2016-02-04
Move ADL and EDM converters from opibuilder.feature to separate features


mfurseman:
At the moment the opibuilder feature contains the following plugins:
- org.csstudio.opibuilder.converter
- org.csstudio.opibuilder.adl2boy
- org.csstudio.utility.adlParser

As far as I know these are only useful to people who have existing EDM or MEDM files, I would like to move them to two separate features, for instance:
org.csstudio.applications.opibuilder.edm.feature
org.csstudio.applications.opibuilder.medm.feature

This would allow people who want these plugins to explicitly include the features in their product configurations.

willrogers:
+1 

In particular in our product we have the 'ADL Parser' and 'ADL-2-BOY' perspectives, and they are not useful to us.

kasemir:
We'd likely include both the EDM and MEDM support, but having them in optional, installable features instead of the primary opibuilder feature is of course fine.

berryma4:
@shroffk, @willrogers, @utzeln, @mjclarke01, @jpenning, @ksdj
Please take note, this merge may require changing your product file.

PR #1599, opened 2016-02-08
adding xygraph to utilities feature

berryma4:
#1595

### DIFF
--- a/applications/appunorganized/appunorganized-features/org.csstudio.utilities.feature/feature.xml
+++ b/applications/appunorganized/appunorganized-features/org.csstudio.utilities.feature/feature.xml
@@ -58,4 +58,17 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.csstudio.swt.xygraph"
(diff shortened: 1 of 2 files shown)


jbobnar:
The problem is even bigger and I think it requires a total rethink and redesign of figures. 

xygraph.rap and xygraph.rcp are both parts of the opibuilder-features. The same is true for their host plugin org.csstudio.swt.xygraph. That is not OK, because that feature combines plugins from different modules. There should be a feature in the appunorganized, which provides the plugins from this module. Even better, there should be several features, each providing a subset, such as xygraph, email, graphene etc. 
Opibuilder-features (and other features) can depend on these features, but should not include any of the plugins that come from a different module than the module where the feature is based in. This has already come up in another issue https://github.com/ControlSystemStudio/cs-studio/pull/1194#issuecomment-110745855 and is also related to this #1594.

I can add the xygraph to appunorganized, but at the same time we should remove it from the opibuilder. The downside is that this might break the product configuration for most of the institutes.

Issue #740, opened 2014-10-02
Macros cleanup


kasemir:
DataBrowser and especially opibuilder use macros.
Should be combined into one with some cleanup.

`org.csstudio.apputil.macros`
Does it have all but "inherit" flag?

`org.csstudio.opibuilder.util.MacrosInput`
- Why does it need to enforce the order of macros? Fundamentally, macros are like environment variables, which are not ordered.
- Needs to behave like a map, but should not expose the actual map.

PR #1600, opened 2016-02-09
Macros

willrogers:
This is not for merging, but for discussion.

#740 discusses cleaning up macros.  There are different implementations in CSS, but the problem they're trying to solve is quite simple.  There is also a new implementation in `org.csstudio.display.builder` and that one looks perfectly good to me.

In this pull request I've created the interface `IMacros` and a simple class `Macros` to implement it.

`MacrosInput` and `MacroTable` now extend this class to avoid duplication.  I have retained the interface `IMacroTableProvider` for compatibility.

I've also changed the `MacrosInput` class to avoid exposing the internal map.

I think @kasemir  would be keen to remove `include_parent_macros` from OPIBuilder at which point we could use a generic version in `org.csstudio.apputil.macros` everywhere.

Thoughts @berryma4 @shroffk @jbobnar ?


jbobnar:
As Kay mentioned in #740 the macros don't need to be ordered. I think you can replace the LinkedHashMap in AbstractContainerModel with a normal HashMap (https://github.com/ControlSystemStudio/cs-studio/pull/1600/files#diff-ed8e71a1517af6e8325b40f9c7250a6cR47).

Macros class is using ConcurrentHashMap. Is that really needed or could you use a normal HashMap instead? If concurrent implementation is needed, than MacroTable probably needs it as well. 

I think it would be nicer if Macros had another constructor Macros(Map<String,String>) and in the MacroTale you could call that constructor rather than overriding the macrosMap. This way you won't need to expose macrosMap at all (keep it private) The same can be done in MacrosInput.

MacrosInput#setMacrosMap is "exposing" the internal object. It would be better to add a method setAll(Map<String,String>) to the IMacros interface, which would do something like 
```
macrosMap.clear(); 
macrosMap.putAll(map);
```

willrogers:
Good point.  I can't find any reference to that method so I have removed it.

willrogers:
You're also correct that it is an error to override the map implementation in subclasses.

kasemir:
I agree that the new implementation in org.csstudio.display.builder is good. That's the general idea of the display.builder development ;-)

As for ordered, when macros are "dumped" in the XML file or in a widget dialog that shows the current macro settings, alphabetical order is good. For macro evaluation it obviously doesn't matter.

willrogers:
Right now it's difficult to share the implementation in Display Builder.

Is Display Builder depending on CS-Studio utilities?  If we had the right implementation in CS-Studio could you use that?

berryma4:
I'm adding @shroffk as a reviewer.  I know BNL heavily uses containers with macros.

willrogers:
I don't mind doing some work on this, but I need some guidance to get it right.

kasemir:
Hi Will:

Yes, the display builder can use generic csstudio plugins like o.c.java or o.c.apputil.
So if you put a "perfectly good" macro support package there, it could be shared by display builder, opibuilder, data browser, who knows what else.

What's now in https://github.com/kasemir/org.csstudio.display.builder/tree/master/org.csstudio.display.builder.model/src/org/csstudio/display/builder/model/macros has these advantages:

* Most basic MacroValueProvider
* Macros implements the MacroValueProvider based on a concurrent hash, has a 'merge' with shortcut if one of the inputs is empty, and getNames() provides them in alphabetical order which is good for constantly saving them in XML or showing them in a dialog
* MacroHandler seems to handle the macro replacement just fine, there are unit tests, haven't found a problem so far

What to watch for:
* The MacroValueProvider in that package is not generic, it's specific to the display builder. So that's just not meant to be copied outside of the display builder
* There's no `include_parent_macros` idea, because I hope it's just not necessary.
* The MacroXMLUtil depends on some other display builder XMLUtil, so those would have to be copied out together. 
* There's _only_ XML support, no other text-based serialization. The opibuilder has its own way of serializing macros for the property files: `macro1,value | macro2,"another value"`. When I started to implement that so that, with tests to assert that it allows including those (escaped) separator chars inside the value, I soon decided that it's a lot of work and much easier to simply stick with one macro serialization, namely XML.

berryma4:
Please merge master to this branch.  A Travis ci yml file was added for testing.

willrogers:
No consensus - I'll close this.

PR #1618, opened 2016-02-11
Suggest changing perspective on opening OPIEditor (4.2.x)

willrogers:
Although this is a new feature, I would like this on 4.2.x so that it gets into our release and there's one less thing to train our users.

It is a squashed commit from the branch merged into master:

Tie to a preference so that choice can be remembered.  Set up listeners
on workbench components to ensure this is active for each new part.

Improve messages in perspective-switching dialog.

Improve logic in perspective checking.

Move OPIEditor-specific logic into constructor.

This shows how this functionality could be made more general by
providing an abstract class in a utility plugin and extending it
in specific plugins.

Remove author tag.

Handle possibility that active workbench window is null.

Fix logic error in PerspectiveChecker.

Always add EditorWindowListener to the current workbench.
Slight restructure to make logic more intuitive.  Add Javadoc comment.

Add default preference for perspective switching.


berryma4:
I agree with this.  I will do another release after this merge (4.2.2), do you have anything else?
@kasemir, @shroffk , @utzeln, @mjclarke01, @charlesHenryPATARD, @ksdj, @xinyuwu, @jpenning
I'm tagging site reps, because of this being an exception to our normal rules.
I'll give a day for objections.

Issue #1573, opened 2016-01-20
PVUtils getLabels()


laurent-PHILIPPE:
I add a method PVUtil.getLabels()in the GANIL CSS distribution.
This method allow to get the possible values of a PV of type enum in css script.

Is it possible to add this method in the next release of CSS?

kasemir:
Sure, create a pull request.

PR #1621, opened 2016-02-16
Issue #1573 Get list of possible enum values

laurent-PHILIPPE:
Link to issue #1573.

Sorry for the delay in pull creation.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java
@@ -18,6 +18,17 @@ import org.csstudio.opibuilder.util.ErrorHandlerUtil;
 import org.csstudio.simplepv.IPV;
 import org.csstudio.simplepv.VTypeHelper;
 import org.csstudio.ui.util.thread.UIBundlingThread;
+import org.diirt.util.array.ListInt;
+import org.diirt.util.array.ListNumber;
+import org.diirt.util.time.Timestamp;
+import org.diirt.util.time.TimestampFormat;
+import org.diirt.vtype.AlarmSeverity;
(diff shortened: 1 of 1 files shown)


willrogers:
Hi,

I'm not sure where it's documented (@berryma4, is it somewhere?) but we have a convention to use four spaces for indentation and no whitespace at the end of lines.

You can make Eclipse show you:
spaces-and-tabs

I've put a pull request to your branch to fix it.  Don't be put off contributing we're all friendly!

Cheers,
Will

laurent-PHILIPPE:
Ok, thanks for the correction.
Is someone proposing an eclipse code formatter to assure a correct formatting?

willrogers:
Yes, this is a good idea if it doesn't exist already.

kasemir:
I use https://github.com/ControlSystemStudio/org.csstudio.sns/blob/master/plugins/org.csstudio.sns.product/code_format.xml to format the code. That gets you the basic space-instead-of-tabs and puts the braces in the correct location.
In addition, I enable Editor "Save Actions" to trim trailing whitespace, organize imports.

berryma4:
Funny, we have the checkstyle and not the format.xml.  Maybe we should move Kay's to cs-studio/build/?

berryma4:
And, thank you @willrogers

kasemir:
For what it's worth, regarding the List to array conversion, there are two useful idioms when converting a list to an array.

1) Let the `toArray()` allocate the memory
```
(String[]) labels.toArray(); 
```
2) You allocate the array with the proper size, `toArray()` only populates it
```
labels.toArray(new String[labels.size()]);
```
in which case the return type already has the correct type.

In Eclipse, if you just type 'toarray' and use completion, you get a template for the second case, which is generally recommended.

The committed code uses
```
labels.toArray(new String[]{}); 
```
which allocates an empty array, passes that to `toArray()` which then needs to allocate another array for the correct `size()`.

No, this extra allocation for naught doesn't matter in the big picture. Just nitpicking.

berryma4:
@kasemir or?
labels.stream().toArray(String[]::new);

PR #1626, opened 2016-02-22
Perspectives

willrogers:
@nickbattam @mfurseman here's the prototype for perspective loading and saving.

Don't merge this - comments first, then we can merge to test locally before suggesting to CS-Studio.


willrogers:
Sorry for making the pull request on this repo - that was a mistake.  I will submit a refactored one soon.

Issue #1574, opened 2016-01-22
Saving perspective doesn't save contents of OPI views


willrogers:
* Open new workspace
* Create new OPI
* Open OPI in workbench
* Save perspective with new name "Test"
* Window | New Window

The new window will open in the Test perspective, but the OPI view will be empty.

I'm puzzling over this a bit.  When you shut down CSS, `saveState()` is called and you can save data to the memento.  Then when you re-open CSS the view is correctly restored.  However, if you save a perspective `saveState()` is not called, and when the perspective is reset no memento is provided to the view.

I don't know the best way to handle this.  @kasemir, do you know about this?

kasemir:
Looks to me like the perspectives are per window.

If I create a perspective in Window 1, I can close, reset, get it on restart.

But in Window 2, such a perspective won't work. The perspective shows up in the list of perspectives, but is deficient as you describe.

You can still create a perspective where you start in window 1, open a bunch of panels, drag some of them out into their own separate window (*), then save that as a perspective
-> Now you do have a saved perspective with multiple windows. As you switch perspectives, those separate windows will open/close depending on the active perspective.

.. But those separate windows (*) are 'detached' windows that belong to the same original workbench window. Only the original workbench has a menu and tool bar, the detached windows associated with it do not have menu nor toolbar.

So bottom line is that I can create several perspectives for a beam line, they can consist of one or more windows which open/close with the perspective, as long as I keep it to one main workbench window. If I use Window|New to open additional workbench windows, the perspectives don't quite work for the new window.

I agree that this is unfortunate, but not sure what to do about it. Also no show stopper, since I can create the various layouts as long as I create separate windows as detached windows, and not as new workbench windows. To users that's actually a bit better, because when we used multiple workbench windows, File/Exit in one would close all of CSS. With just one workbench window (and menu and toolbar), the other windows being detached windows (no menu, no toolbar), there is only one File/Exit, and there's no surprise that it indeed closes CSS.

willrogers:
I've looked at this for a while and it is a pain in the bottom.

If you ever open another window, then close your original window first, you can't get back to a working version of your saved perspective.  As you say, it is a perfectly valid workflow to use only one window, but right now the default workflow for editing OPIs introduces multiple windows, because pressing the "play" button in OPI editor launches the running OPI in a new window.

Also, this may present problems when we come to look at sharing perspectives if we want OPIs in them.

willrogers:
[...]

Issue #1398, opened 2015-10-20
Save and share perspectives


willrogers:
This was one of the original goals of #544, but all other aspects of that ticket are complete.  Using views for OPIs enables more control over perspectives.

Kay's original description:

> Goal is to be able to “save” the layout of *.opi displays, so it can be “restored” and operators can switch between different layouts.
> Maybe including a “frozen” mode where a “restored” layout cannot be altered.

Developers at Diamond are keen to be able to share perspectives between different workspaces.

willrogers:
Copied from #544, by @kasemir:

Some clarification what I think this is about:

> Users want to save and restore the layout of *.opi panels on the screen.
> As an example, the current layout may be called "Vacuum". "VacPumps.opi" and "VacValves.opi" are open in some arrangement on the screen.
Then operators can switch to a "Cooling" layout that instead opens "WaterPumps.opi" and "WaterTemperatures.opi" in some other arrangement on the screen.
Operators can switch between the two layouts.

> I still think that these layouts can just be the RCP Perspectives, as also suggested by #438. Changing the opibuilder.Runtime to exclusively use "Views" instead of "Editors", and fixing how those "Views" keep their *.opi file path (#674) should mostly get us there. Unclear if perspectives would be configured by adding some new XML format file reader as per #438, or we study how perspectives are currently persisted and somehow create those files, or only use "Save As" and then copy the persisted perspectives from a development computer to the control room.

> "Frozen" has been brought up in that operators should not be allowed to move things around or close windows. The RCP API for the Perspective offers IPageLayouts.setFixed(true), but that actually no longer functions in E4 (https://bugs.eclipse.org/bugs/show_bug.cgi?id=331402). Realistically, I consider this a weak requirement. EDM etc. never prevented window movement. The web or email browser in which you read this can be moved. Operators will certainly need to open a Data Browser, start a logbook entry, maybe move windows around to better see what they need to see.
> A more realistic requirement is a "Reset": In case operators moved windows or for example closed the "WaterPumps.opi", a "Reset" of the "Cooling" perspective should restore it to the original layout, re-opening the "WaterPumps.opi". Fundamentally, perspectives already support "Reset", so that should not be a technical problem.

willrogers:
This Eclipse bug is related: https://bugs.eclipse.org/bugs/show_bug.cgi?id=378811

willrogers:
Copied from #544, by @kasemir:

### Snapshot, Restore complete Workbench

[...]

PR #1627, opened 2016-02-24
Perspectives

willrogers:
Don't merge!

This is a prototype method for importing and exporting perspectives **including** OPIViews.  There's a lot to talk about here so I thought I could open a pull request for discussion.

This depends on functionality that is due for release on Friday: https://www.eclipse.org/eclipse/development/plans/freeze_plan_4_5_2.php.

Most of this is in a separate plugin / feature, so can be included or not by individual sites.  There is some extra code in `OPIView.java` needed to restore state in some cases.

See #1574 and #1398.

@kasemir @shroffk @berryma4


kasemir:
This would be huge!!
Can we have a hangout where you explain/demo?
Friday, March 26, the usual CSS time of 9:00 Eastern US, 14:00 London?

shroffk:
March 26th or Feb 26th?

willrogers:
I have a meeting on Friday at 14:00.

11:00 Eastern / 16:00 London would be ok.

kasemir:
I'm OK with Feb 26, 11:00 Eastern / 16:00 London

berryma4:
I'm in!

kasemir:
I'll create hangout invite

kasemir:
Notes from hangout:

"Perspective" only contains information for "opiView:1234", "opiview:3455".
The MPart info has the state, i.e. the *.opi path, macros etc. for each of these views.
So when just the "Perspective" is saved, we get a blank "opiView:1234" and "opiview:3455" but they lost their state.

Will moved the 'state' into the MPerspective, so it'll be saved/restored.

Eclipse 4.5.2 does save & load the perspectives as part of a load/save preferences mechanism using *.epf files. Eclipse does _not_ have a "save this perspective" or "load that perspective" mechanism for individual perspectives.

So Will extended the "Save As" Perspective to also store that one perspective as a file. For now that's done silently in the workspace (data) area. User is not prompted for path, no way to change the path.

In addition, there's a new menu entry CSS/Utility/Load Perspective to load just one such file.

Will demo'ed it, works great.

--> We can load & save selected perspectives.
--> Needs 4.5.2 to actually "work". Our pom.xml point to 4.5, which includes 4.5.2 since that's become available.
--> Means that we will have to support this as Eclipse changes.

kasemir:
Idea by Kunal:
On startup, load all saved perspectives from a directory.

willrogers:
@kasemir @shroffk where should perspectives be loaded from at startup?  Should this be saved in a preference?

berryma4:
Could we follow the others, a preference with the default being in the
configuration directory?

willrogers:
Which others are you thinking of?

berryma4:
Diirt, then the common product also has the preference.ini, and jaas
files.  I move these for debian (etc/cs-studio), but it's a nice default.

kasemir:
A default location inside the product makes sense as a default, because that way you can package your product with some "built in" perspectives that everybody gets with the product download, and they're then available regardless of workspace.

For my production setup on beamlines, however, the actual location will be outside of the product.
Several versions of CSS will all need to point to some common directory for the perspectives.

The workspace seems the least useful place.

==> Default to configuration area, with preference to point it to any directory.

willrogers:
OK there's now a preference option and perspectives will be loaded from that location at startup.

willrogers:
I haven't tested on Windows and I haven't tested with HTTP OPIs, because these are both a bit less convenient for me.

willrogers:
@berryma4 @kasemir there is already a perspectives preference page in "General | Perspectives".

I added a "CSS Core | Perspectives" preference page.

It's definitely confusing to have them in different places, but possibly confusing to have the CSS one in the general section.

Should I move my preference page to "General | Perspective Loading"?  I think so.

berryma4:
Tough question, but I agree with "General | Perspective Loading"

willrogers:
OK - bar further unit tests and any comments people have, I think this branch is ready.

I'll resolve the merge conflicts tomorrow, then either you can:

* wait for further comments
* wait until someone else builds and tests it 
* merge this and we fix any further problems we find on master

willrogers:
A user comment from here is that we need to be able to configure both:

* location from which perspectives are loaded
* location to which perspectives are saved

I'll add the former.

willrogers:
OK I think this is ready to merge.

berryma4:
This is great!!!  Thank you!

Issue #1628, opened 2016-02-24
PostgreSQL archive reader exception


jbobnar:
The following exceptions happened occasionally.
```
org.postgresql.util.PSQLException: ERROR: portal "C_8" does not exist
  at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2198)
  at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:1927)
  at org.postgresql.core.v3.QueryExecutorImpl.fetch(QueryExecutorImpl.java:2130)
  at org.postgresql.jdbc2.AbstractJdbc2ResultSet.next(AbstractJdbc2ResultSet.java:1917)
  at org.csstudio.archive.reader.rdb.RawSampleIterator.next(RawSampleIterator.java:154)
  at org.csstudio.trends.databrowser2.archive.ArchiveFetchJob$WorkerThread.run(ArchiveFetchJob.java:140)
  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
  at java.lang.Thread.run(Thread.java:745)
```
```
org.postgresql.util.PSQLException: Cannot change transaction read-only property in the middle of a transaction.
  at org.postgresql.jdbc2.AbstractJdbc2Connection.setReadOnly(AbstractJdbc2Connection.java:741)
  at org.csstudio.archive.reader.rdb.RDBArchiveReader.<init>(RDBArchiveReader.java:106)
  at org.csstudio.archive.reader.rdb.RDBArchiveReader.<init>(RDBArchiveReader.java:81)
  at org.csstudio.archive.reader.rdb.RDBArchiveReaderFactory.getArchiveReader(RDBArchiveReaderFactory.java:48)
  at org.csstudio.archive.reader.ArchiveRepository.getArchiveReader(ArchiveRepository.java:111)
  at org.csstudio.trends.databrowser2.archive.ArchiveFetchJob$WorkerThread.run(ArchiveFetchJob.java:126)
  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[...]

PR #1629, opened 2016-02-24
Postgres Archive issues

jbobnar:
Pull request fixes the issue #1628

@matthieulambertiter can provide more details about the fix.

### DIFF
--- a/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/META-INF/MANIFEST.MF
+++ b/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: RDB Archive Reader
 Bundle-Description: Support for reading archived data from Relational Database (Oracle, MySQL)
 Bundle-SymbolicName: org.csstudio.archive.reader.rdb;singleton:=true
-Bundle-Version: 4.2.0.qualifier
+Bundle-Version: 4.2.1.qualifier
 Bundle-Vendor: Kay Kasemir <kasemirk@ornl.gov> - SNS
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Require-Bundle: org.eclipse.core.runtime;bundle-version="3.6.0",
--- a/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/pom.xml
+++ b/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/pom.xml
@@ -8,6 +8,6 @@
     <version>0.0.1-SNAPSHOT</version>
   </parent>
   <artifactId>org.csstudio.archive.reader.rdb</artifactId>
-  <version>4.2.0-SNAPSHOT</version>
+  <version>4.2.1-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 </project>
--- a/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/src/org/csstudio/archive/reader/rdb/ConnectionCache.java
+++ b/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/src/org/csstudio/archive/reader/rdb/ConnectionCache.java
@@ -56,6 +56,11 @@ public class ConnectionCache
                    user.equals(other.user) &&
                    Objects.equals(password, other.password);
         }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(url,user,password);
+        }
     }
 
     /** Cached RDB connection with reference count */
@@ -144,5 +149,16 @@ public class ConnectionCache
         }
         entry.rdb.close();
         logger.log(Level.FINE, "Closed {0}", entry.id.url);
+    }
+
+    /**
+     * Clear cache content
+     */
+    public static void clean()
+    {
+        synchronized (cache)
+        {
+            cache.clear();
+        }
     }
 }
--- a/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/src/org/csstudio/archive/reader/rdb/RDBArchiveReader.java
+++ b/applications/archive/archive-plugins/org.csstudio.archive.reader.rdb/src/org/csstudio/archive/reader/rdb/RDBArchiveReader.java
@@ -102,13 +102,16 @@ public class RDBArchiveReader implements ArchiveReader
         this.use_array_blob = use_array_blob;
         timeout = RDBArchivePreferences.getSQLTimeoutSecs();
         rdb = ConnectionCache.get(url, user, password);
-        // Read-only allows MySQL to use load balancing
-        rdb.getConnection().setReadOnly(true);
+
 
         final Dialect dialect = rdb.getDialect();
         switch (dialect)
         {
         case MySQL:
+        	// Read-only allows MySQL to use load balancing
+        	if (!rdb.getConnection().isReadOnly()) {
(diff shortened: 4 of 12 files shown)


berryma4:
Thank you!
@kasemir okay with you?

kasemir:
Where do you actually enable or disable the 'concurrency'? Is that only done in a Would it be easier to always enable/disable for Postgres, or make that a preference setting? That way, it would only be checked in `RawSampleIterator`, instead of having to pass it down all the method calls.

As for `displayUnknowChannelException` (Unknow_n_?) and the existing `prompt_for_errors` preference setting, the existing logic doesn't work in our setup. Sure, users would like to know if there is no archived data for a channel. Not a big problem, but good to know if there's only live and no historic data because
a) There's a problem
b) That channel simply isn't in the archive
But our setup includes at least 2 archive data sources: One for the accelerator, and one for instruments. So even archive resp. instrument channels that are archived get the warning because of the 'other' data source that doesn't have the channel. The overall logic thus needs to check:
If channel is not found in _any_ of the sources, then warn about the unknown channel.

jbobnar:
ITER extended the ArchiveFetchJob and they set both preferences via the constructor. I don't know what conditions trigger the error, because I wasn't involved in this.

The displayUnknowChannelException is true by default (it can only be changed if the protected constructor is invoked with a proper parameter), so nothing changes for you. Whenever the exception happens, you will still be notified the same way as you were before.

kasemir:
> ITER extended the ArchiveFetchJob and they set both preferences via the constructor. 
.. which is why those new arguments need to be passed down many calls.
Wouldn't it be easier to just use the database dialect or a preference inside the RawSampleIterator, so no need to change constructors and all the method calls down?

> displayUnknowChannelException is true by default ..nothing changes for you
Right. It's adding new code without doing anything about the real problem: The determination of "unknown channel" is wrong. Should check if channel unknown in all data sources, i.e. not found in any data source, instead of complaining N-1 times if in N data source one has the channel so all is fine.

So bottom line: Sure, we can merge this because it doesn't cause any harm. Still, could it be rewritten to be overall better?

jbobnar:
> Wouldn't it be easier to just use the database dialect or a preference inside the RawSampleIterator, so no need to change constructors and all the method calls down?

Not if they need one option in one case and the other option in another case. They have a BOY widget, which also fetches the history data. I don't know what happens inside that widget that causes the exceptions listed in #1628, but they only happen when the widget is used. When using the databrowser the old way still works fine.

I haven't tested this, but considering what those additional parameters to the ResultSet do, I am guessing that the performance drops when they are used, because the result set needs to be in sync with the DB. Therefore, it makes sense to use the old way when possible and the concurrency only when really needed.

> Still, could it be rewritten to be overall better?

displayUnknownChannelException case is similar. In databrowser they want to have the same behaviour as you do, but in the widget, they don't care about missing channels. 

However, the decision whether to show the error or not should actually be in the listener and not in the job. I've changed that in d8cc97b.

berryma4:
Status?

kasemir:
I'd summarize the status as this:
Adds a new 'concurrency' parameter for ITER's use of Postgres. Unclear to me why this option needs to be passed down instead of being a preference setting. Also unclear to Jaka why only the BOY widget experiences the problem, and not the data browser. So overall fishy, but since it doesn't seem to break anything, and considering that ITER really wants this
--> What the heck, just merge it

Original code change also touched on the handling of unknown channels. We could have expanded that to really handle the problem by properly checking if a channel is unknown in _all_ archive data sources, which might be reported to the user, or if a channel is known in at least one data source, so nothing should be reported to the user. In the end, any reference to unknown channel handling has been removed from the pull request.

jbobnar:
@kasemir  I see that in the case when there was an error fetching the data, the archive source is removed from the PVItem, but only if doPromptForErrors is true. See [here](https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/ui/Controller.java#L164). Is this a bug? Shouldn't this happen every time when unknown channel exception is thrown? 

If the source should be removed from the item every time, than it make sense to report every unknown channel exception. Otherwise, I can update the code to not report the exception if the channel was found in at least one archive source.

kasemir:
These are the use cases:

1) Just one archive data source. User added a channel that simply doesn't have archived data.
The first time we try to fetch data, with doPromptForErrors=true, user gets the error message.
User clicks "OK", and can from now on zoom, pan, .. without getting that same error again.

2) Several archive data sources: One for beam lines, one for the accelerator, maybe one with older data. User added a channel that has archived data, but only in one of the data sources. Unfortunately, user still gets an error for the 2 other data sources, even though everything is OK. Still, user only gets that bogus error _once_ per data source, and can from then on zoom/pan without seeing the same error again.

So the feature to only report the error once is important. In use case 2, there should actually be no error reported at all. There are several ways to implement the 'at most once'. Could have additional flags to track if the error had already been reported. Simply not checking the archive data source which doesn't have the error "works", plus has the advantage that it's faster because we no longer bother to look in an archive where we already know the channel can't be found.

kasemir:
As for always removing the data source, even with doPromptForErrors=true: You're right, that would be OK, as long as we still have a log message.

jbobnar:
Implemented handling of unknown channel exceptions like you proposed in the previous comment.

One thing that I am not 100% is [this ](https://github.com/ControlSystemStudio/cs-studio/pull/1629/files#diff-92c5613716d06ba6e1f1d030e6355feeR163). It's clear that we should remove the archive source in case of unknown channel. But should we also remove the source for any other exception? Most likely yes, but just in case think about all the use cases that can cause an exception (e.g. DB server unavailable, or timeout or any kind of runtime exception etc.)

kasemir:
Jaka, thanks for only reporting the not-found if not found at all, not in any of the sources!

As for removing the source response to any error: I think that's fine. People don't need to be bothered again and again for the same error. If it's a database error that later recovers, you need to close/reopen that *.plt file.
Removing the source was an easy, practical way to only bother the user once, plus it avoids loosing time querying the same source again. The alternative would be an option "Don't show this message again" in the error dialog, then hide the message, but still query the source because maybe later it'll work out. Not sure if that's really better than the simple remove-source solution.

jbobnar:
@berryma4 I think this is all done and can be merged. 
@kasemir, please raise the flag, if that is not the case.

berryma4:
We seem to have an issue:

https://openepics.ci.cloudbees.com/job/cs-studio-applications-databrowser-master/61/
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:0.23.1:compile (default-compile) on project org.csstudio.trends.databrowser2: Compilation failure: Compilation failure:
[ERROR] /scratch/jenkins/workspace/cs-studio-applications-databrowser-master/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/archive/ArchiveFetchJob.java:[126]
[ERROR] the_reader.enableConcurrency(concurrency);
[ERROR] ^^^^^^^^^^^^^^^^^
[ERROR] The method enableConcurrency(boolean) is undefined for the type ArchiveReader
[ERROR] 1 problem (1 error)
[ERROR] -> [Help 1]
```

berryma4:
Never mind.  This was a Jenkins issue.
I apologize.

jbobnar:
The core is there
https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/archive/archive-plugins/org.csstudio.archive.reader/src/org/csstudio/archive/reader/ArchiveReader.java#L183

I think it is the "chicken and egg" problem again. The archive module should be built first.

PR #1649, opened 2016-03-08
Fix checkstyle warnings

nickbattam:
@berryma4 

Final push to get the checkstyle warnings down to manageable levels (core: 29, applications: 69)

The remaining 'missing @Override but has @InheritDocs' warnings are the result of one of two reasons:
1 - the plugin specifies Java1.5 in which @Override was invalid for an interface but the checkstyle rules are assuming a newer Java (inc de.desy.language.*)
2 - the method is not overriding anything. This could be the result of change of interface/base class (a genuine bug), or simply a cut+paste comment error

I didn't want to impose arbitrary resolutions in either of those cases. Hopefully someone who know more about files/modules containing the remaining warnings can resolve them appropriately.


berryma4:
Thank you again so much!

Issue #1644, opened 2016-03-04
PVTable breaks the build


jbobnar:
@charlesHenryPATARD 
The latest changes in pvtable broke the build. There are some references to classes Configuration and Measure, but the classes are nowhere to be found. Probably you forgot to add them to git.

I would also like to ask you to reduce the version number on the org.eclipse.core.runtime to 3.11.0, or remove the version number completely. ITER is locked to eclipse mars (without service pack), and it is too much hassle to upgrade to SR1. Thank you!

jbobnar:
Several action classes are also missing.

berryma4:
#1587

berryma4:
See Jenkins for build errors:
https://openepics.ci.cloudbees.com/job/cs-studio-applications-display-master/44

nickbattam:
This merge added 705 new checkstyle warnings.
A quick look suggests they are all from tabs and trailing whitespaces.

berryma4:
this merge (#1647) added more checkstyle errors

jbobnar:
I get the following error when adding the pvtable to the product

```
!ENTRY org.csstudio.display.pvtable 4 0 2016-03-08 16:27:32.816
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.csstudio.display.pvtable [654]
  Bundle was not resolved because of a uses contraint violation.
  org.osgi.service.resolver.ResolutionException: Uses constraint violation. Unable to resolve resource org.csstudio.display.pvtable [osgi.identity; osgi.identity="org.csstudio.display.pvtable"; type="osgi.bundle"; version:Version="4.3.1.qualifier"; singleton:="true"] because it is exposed to package 'org.apache.poi.ss.usermodel' from resources org.apache.poi [osgi.identity; osgi.identity="org.apache.poi"; type="osgi.bundle"; version:Version="3.13.0"] and org.apache.poi.ooxml [osgi.identity; osgi.identity="org.apache.poi.ooxml"; type="osgi.bundle"; version:Version="3.13.0"] via two dependency chains.

Chain 1:
  org.csstudio.display.pvtable [osgi.identity; osgi.identity="org.csstudio.display.pvtable"; type="osgi.bundle"; version:Version="4.3.1.qualifier"; singleton:="true"]
    require: (&(osgi.wiring.bundle=org.apache.poi)(bundle-version>=3.13.0))
     |
[...]

Issue #1515, opened 2015-12-11
PVTable Updates


kasemir:
From Charles-Henry PATARD:
![image004](https://cloud.githubusercontent.com/assets/1932421/11745177/2816cdb0-9fe4-11e5-9f18-549895544b14.jpg)

We finally finished our PVtable coding. We’re ready to propose you new PVtable functionalities. So, just below a short overview :
-        Table export to Excel file
-        New column Timestamp Save which save the timestamp for each snapshot
-        Ability to create a configuration list of PVs if a line start with #conf#
-        If a configuration list exists in the table the user can  “make a measurement”. Make a measurement is the ability to take a snapshot of the configuration PV list and add the snapshot at the end of the table.
-        The user can delete the last measure or all the measures.
The native behaviors of PVtable still available.
 
We committed on master branch of cs-studio git repo. We would like to know if those features are good and useful for cs-studio project ?
So, now I would like to know how to create a pull request ?

kasemir:
You describe several features.

Table export to Excel is good, we should merge that.

New column TimestampSave that holds the time stamp of the saved value is good. Except for maybe labelling it "Saved Value Timestamp", I'm for merging that. In principle, the visibility of that column could be configurable just like the "show_description" and "show_units" columns which were added but can be enabled/disabled via preferences. I'm leaning towards always showing it, though, since the time when a value was saved should be generally useful.

As for the configuration list of PVs, I'm afraid I don't understand how that works and what it's good for.
Showing the same PV several times within the same table, with different "measurement" values, clashes with the original design of the table which takes a snapshot of PVs and then allows you to restore all of them, or selected ones. Restoring all of them would be very confusing if the same PV is listed multiple times with different values.

kasemir:
Forgot to answer your question on how to create a pull request.
We use the basic github workflow, see for example https://help.github.com/articles/using-pull-requests/,  https://help.github.com/categories/collaborating/.

Basically:
1) You "fork" the repository via the "fork" button on the github page.
2) You can then fetch a clone of that fork, change the sources in there.
    You could do that on "master", but typically we use some branch like "issue_1515" that includes the ticket number.
3) When done, create a pull request from your_fork:issue_1515 into cs-studio:master

If the pull requests come in small pieces like "Excel table export" where we've already agreed on the ticket that it's a good idea, it's quickly merged.

berryma4:
some of this overlaps with save restore:
[...]

PR #1652, opened 2016-03-09
move apache poi to servicemix bundle of poi

berryma4:
#1644


berryma4:
I moved apache poi to servicemix bundle of poi

berryma4:
from project #1515

PR #1659, opened 2016-03-14
Cleanup tabs and trailing whitespace

nickbattam:
Remove tabs and trailing whitespace from the xml and html files introduced with #1515


berryma4:
Thank you!

PR #1664, opened 2016-03-14
Pvtab egu 1625

### LINKED ISSUE #1625: PVTable 'EGU' handling
The PV Table displays values with units, for example "3 mm".
It also allows entering a new value in that value cell, but that entered value must NOT include the units because the string is currently written as entered.

Ideally, the PV table would know that it had added the " mm" units and remove them when sending an entered value back to the PV.

kasemir:
Fixes #1625


kasemir:
Wait: Will also fix
```
Warning: NLS missing message: InformationPopup_DelConfHeader in: org.csstudio.display.pvtable.messages
```
in this pull request

kasemir:
OK, added texts that had only been in the Fench version.

berryma4:
done?

kasemir:
Yes.

Issue #1661, opened 2016-03-14
Minor AlarmTable Issues


jbobnar:
The following issues were observed:

- Some linux systems don't handle tooltips very well. If it happens that tooltip goes over the screen boundaries, the OS tries to move the tooltip inside the boundary. At the same time it somehow recognize that as mouse move, which causes the tooltip to disappear.
- Alarm Table should use the time format from the preferences (if it exists) for all tables (including the primary table).
- If the header is suppressed and there is no connection to the server, the tables are not disabled (this is visible in the alarm table opiwidget).
- if the preferences don't define a color pair for MinorAck/MajorAck/InvalidAck severities, their non-ack severity color should be used. This is only true when SeverityColorPairs are used.

jbobnar:
I will hijack this ticket to add two more issues. They both appear only when using the alarm table in opis:
*  If the alarm table model is in "timeout" state when the edit part is created, the table does not show the "disconnected" border.
* If no column is selected in the column configurer, the OK button sometimes becomes disabled.

PR #1669, opened 2016-03-17
Alarm table bugfixes

jbobnar:
This fixes the two additional issues listed in #1661 (https://github.com/ControlSystemStudio/cs-studio/issues/1661#issuecomment-197284438):
* the alarm table border is incorrect, if the alarm model already exists when the table is shown,
* column configurer sometimes disables the OK button if no column is selected in the list.

I also replaced several listeners with lambdas.


berryma4:
@kasemir okay?

kasemir:
fine

Issue #1655, opened 2016-03-11
Handling PV-Name prefixes for archive data sources


MichaelRitzert:
I recently switched from the RDB archiver to the archiver appliance, and I find that my databrowser plots are no longer working. The obvious reason is that in the old setup, I had the PVs stored as "ca://name", while the appliance only likes "name". After the switch, this means that I cannot have a plot that can read data from both sources transparently.

I don't really like the situation with the appliance, because our OPIs pass around macros containing the common "ca://${USER}:etc" part of PV names, so I cannot easily create a parametrized plt file – the ca:// would always be there. Just omitting the prefix from the OPIs and relying on the default (or using ca://${name} in every widget) doesn't look to attractive to me.

But also for the RDB archiver, it can be confusing for the users to not find a PV because they forgot the ca://.

Would it be possible to extend the logic of the archive reader(s) to retry without the PV name prefix if a channel is not found?

kasemir:
At earlier times I would have suggested that you drop the ca:// (or epics://) prefix, but there are EPICS sites which start to use both Channel Access and PV Access, so their displays as well as archive configurations will require some prefixes.
For example, the default may be ca:// but some PVs now use pva://. Or you switch the default to pva://, and use ca:// for older channels.
Or you don’t depend on any default and always specify ca:// resp. pva://

In any case, I think this is best handled by the archiver appliance. Its data server needs to allow prefixes.

I recently switched from the RDB archiver to the archiver appliance, and I find that my databrowser plots are no longer working. The obvious reason is that in the old setup, I had the PVs stored as "ca://name", while the appliance only likes "name". After the switch, this means that I cannot have a plot that can read data from both sources transparently.

I don't really like the situation with the appliance, because our OPIs pass around macros containing the common "ca://${USER}:etc" part of PV names, so I cannot easily create a parametrized plt file – the ca:// would always be there. Just omitting the prefix from the OPIs and relying on the default (or using ca://${name} in every widget) doesn't look to attractive to me.

But also for the RDB archiver, it can be confusing for the users to not find a PV because they forgot the ca://.

Would it be possible to extend the logic of the archive reader(s) to retry without the PV name prefix if a channel is not found?

jbobnar:
> In any case, I think this is best handled by the archiver appliance. Its data server needs to allow prefixes.

[...]

PR #1670, opened 2016-03-17
Archiver Appliance and schema

jbobnar:
Strip the protocol schema from the name before fetching the data. 
Solves #1655


berryma4:
Thank you!

kasemir:
See discussion on #1655. I don't think this really solves the issue. Going forward, we need to allow both "ca://fred" and "pva://fred" because both can be on the network, both are valid, and both are archived.

If the archive appliance is for now limited to ca:// (is it? I thought it uses the PVManager?), then it's OK for now to strip the ca:// from PVs in the appliance reader, and report an error on any other prefix. The plan should then be to support both ca:// and pva:// in the future. Simply stripping the prefix and effectively returning ca://fred when asked for pva://fred seems wrong.

jbobnar:
Appliance is not using pvmanager. It uses the (modified) thin PV layer, that used to exist in CSS before pv manager:
https://github.com/slacmshankar/epicsarchiverap/tree/master/src/main/org/epics/archiverappliance/engine/pv
Adding a ca://fred or pva://fred will result in forever disconnected PV.

> See discussion on #1655. I don't think this really solves the issue.

It depends on whether ca://fred and pva://fred is the same thing or not. If the PV is uniquely defined by its name (doesn't matter if it is pva or ca) than the clients that request data from the archiver do not need to know how the archiver obtained the data. They know that the server has the PV 'fred', which is of type 'X'. If they archiver gathered the data via channel access, pvaccess, http, tango or tine doesn't really matter to the client.

kasemir:
Well, there is a EPICS_V4_PV as well as EPICS_V3_PV in that PV layer, so clearly a motion has been made towards supporting both protocols. Maybe it plans to distinguish between the two by means other than a ca:// resp. pva:// prefix, then it'll be up to the CSS appliance.reader to map the prefix to whatever the archive appliance uses to fetch the corresponding data.
Simply stripping the prefix may well be a good workaround for now, but it can't be the long term solution.

Issue #1604, opened 2016-02-09
Add "class" property to all BOY widgets to make BOY scheme more useful


berryma4:
The idea would be to use the class property much like in html.
In the BOY scheme file, you can have multiple Label widgets with classes; header, input-label, etc.
Or text update classes could be; power-supply, readback, etc.

We could, like html/css allow multiple classes per widget, with each overriding and adding properties.

PR #1671, opened 2016-03-17
Widget Class Implementation

jbobnar:
Related to #1604 

Perhaps this should go into a separate branch before it is confirmed. I think it is all backwards compatible, but just in case try with some of your opis first.

There is a new property called widget_class. Its value can be either a string, a pv or a formula.

First thing you should do is create/edit the BOY Schema opi. Add a few widgets of the same type and define a different widget class for each of them. Change some (visual) properties of those widgets, so that you will be able to see the difference. Save the schema and reload it (restart CSS or via preferences dialog).

Now create a new opi file and add a widget of the type, which you previously added to the schema. Set its widget_class property. The value can be either a string (literal), a PV, or a formula. If your schema contains widget classes 'title' and 'superTitle' , you can do:
- widget_class: title
- widget_class: loc://MyTitleClassPV
- widget_class: ='loc://somePV' > 5 ? "title" : "superTitle"

When you open the OPI (runtime or edit mode), the widget class value (whatever it evaluates to), will be search for in the schema. If a widget of the same type and widget_class is found, "all property values" of that schema widget will be set on your widget. If the widget class is not found, nothing happens.

I quoted "all property values", because some properties are never touched (no matter what) and you can define rules for the others. The properties that are locked are:
- widget_class
- widget_class_value
- widget_type
- widget_uid
- connector src_path, src_term, src_wuid, tgt_path, tgt_term, tgt_wuid

If you didn't specify any rules file, than the following properties will also be ignored:
- name
- x, y position

The rules are defined in a .def file. You can select the def file to use via preferences. I put an [example](https://github.com/Cosylab/cs-studio/blob/widgetClass/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/widgetClassRules.def) into the opibuilder plugin. The details how to define rules are in the example file. I imagine you'll want to set the rules for properties like pv_name, trace_[0-9]+_y_pv etc.

You've mentioned that ideally the properties controlled by the widget class shouldn't be persisted. That would be great, but the problem I have with that is how to modify the schema file. The widgets in schema file have the widget_class defined, but their properties should always be persisted. 
We will either need to add another property which will define the widget classes in the schema file (which I really don't like, because it would only be used in the schema and might be confusing for the users),  or abuse an existing property (like name) to define the widget classes.

Here are some dome opi files as well:
[demo.zip](https://github.com/ControlSystemStudio/cs-studio/files/177998/demo.zip)


jbobnar:
Got another idea how we could solve the persistence issue. What if the schema defined the widget classes with brackets? For example: widget_class: < title >

When saving the file, if the widget class is in brackets, save all properties; if the widget class is not in brackets, skip those properties which are controlled by the class.

kasemir:
I guess the point is that a schema file should _define_ the widget classes and not _use_ them.
Can the code which applies the class properties simply be skipped for the schema file?
A schema file could be identified by
a) matching the current schema preference, but then you can only prepare one schema file
b) `*schema*` in the *.opi file, which would allow you to edit several schema files, no matter if one of them is currently selected via the schema file preference.

jbobnar:
I thought of option a) and shot myself in the foot a couple of times. Option b) is more appealing, but it is a rule that users have to be careful about. If you mistype the name, you lose everything. The XMLUtil would also need to be modified in a way that whenever the widgetModel is persisted, the destination path/name is also provided. That is not very nice.

kasemir:
To a user, having to remember adding "<>" to each class, only doing that in the scheme, but doing it for every class in the schema, sounds harder than simply being aware of "*schema*" in the file name.

Yes, the 'mode' (is this a schema?) somehow needs to be passed to the XMLUtil etc. Maybe not by passing the path/name, but certainly somehow. Can that info be kept in the display model? Or added to the top of the XML document, so even low-level XMLUtil calls can go up the DOM to look for the 'schema' mode?
In any case, handling that in the code seems better than leaving it to the user via a special syntax in the class name.

jbobnar:
@berryma4 Do you have any preferences regarding how to save the opi files? See the above discussion with Kay.

berryma4:
from @kasemir 
> I guess the point is that a schema file should define the widget classes and not use them.

This makes it sound like, a better way to do this is to have a file more like a css file, not a schema.opi that is a opi file itself.

```css
.someHeaderClass {
    font: Header1;
    foreground color: red;
}
```

berryma4:
```css
.className
#widgetName
WidgetType

# ie.
text_update.header1 {
    font: Header1;
}
```

kasemir:
Yes, a "class" (css) file would work.
The basic format is by now also common knowledge, considering that writing such a file is not meant for basic display editing end users but those who install/maintain CS-Studio at a site.
The main downside would be the need for a new parser, since it can't use the existing *.opi file reader.

jbobnar:
> '# = widget name'

We only have the widget class, there are no rules based on widget names. I think that would be confusing for the users: should I use the widget name or the widget class property? Widget names are given automatically by opibuilder and are sometimes used for other things as well (e.g. border caption). There is also no autocomplete for the widget_name.

In general, I am not a big fan of css, because it can be difficult to manage as it grows and can keep you busy chasing ghosts if you mistype a property name (or value). I also don't imagine how to specify complex properties, such as rules or scripts. Even fonts are special cases, because you have the font definition name, or you have family-size-style groups (at the moment the opi file has both). Color properties are similar.

There are some css parsers already available, but getting all this done correctly is larger than our initial estimate.

jbobnar:
I looked around for available parsers.

[jStyleParser](http://cssbox.sourceforge.net/jstyleparser/) seems like the best option (most robust) as it doesn't choke on every minor problem in the css file and is still actively supported. However, complex properties (as mentioned in previous post) are still a problem. And one would need to invent some kind of extension mechanism, because it doesn't support the @extend directive.

JavaFX also has a parser, but is not part of the API.

jbobnar:
I looked at the style sheets a bit more and these are my findings and proposal.

There are some properties, which could never be handled by css, because they are too complex. To handle them you would need to write your own css parser, which should know how to handle those special rules. Those properties are:
- actions 
- scripts
- rules

In addition there are several properties that require special solutions. First you'd need to come up with the syntax how to define them in the css file and then write the parser to convert those values to appropriate structures. Those properties are:
- matrix (symbol widgets)
- stringtable (table widget)
- complex data (scale options property)
- macros 

If any of the above properties are needed, they would require a lot of additional work, which was not in the scope of our contract. 

If you do not care about these properties, I may be able to do something. Below is my proposal, but before I implement anything, I need your confirmation.

```css
/*Defines the general set of properties applied to all widgets of the given type.
The WidgetType is the simple (short) widget typeId - this is the last segment
(after the last dot) of the widget typeId. Dots cannot be part of the WidgetType,
because that violates css rules. We also cannot use the widget_type property, 
because in many cases it contains spaces. */
WidgetType {
  background: someColor;
}
/*Defines the properties for the widget class of the given WidgetType.*/
WidgetType.WidgetClass {  
}
/*Defines the properties for the widget class regardless of the widget type. 
It is valid for text field, gauges, charts etc.*/
.WidgetClass {
}
/*Defines the properties for multiple widget classes*/
.WidgetClassA .WidgetClassB WidgetType.WidgetClassC{
}
/*Defines the WidgetClassD, which has everything what is applicable 
for the WidgetClassA + whatever is in this block */
.WidgetClassA > .WidgetClassD {
}
```

Widget names are not supported. As I mentioned in the previous comment, the name is sometimes used for other purposes and can be misleading. One can always define another class if needed.

Fonts and colors can be defined like this:
```css
@font-face {
    font-family: myBoldFont;
    font-weight: bold;
    font-size: 15;
    src: Tahoma;
}
* {
   lightblue: #99ccff;
   lightgray: #e6e6e6;
}
```
And used in a widget class:
```css
.WidgetClass {
   background: lightblue;
   font: myBoldFont;
}
```

berryma4:
This sounds great (and solves the issues of the schema file).  @shroffk @kasemir thoughts?

willrogers:
This will take a bit of thinking.

Can a widget have more than one class?  And are the available classes selectable, or would you have to type them in?

I prefer CSS to extending the schema OPI.

jbobnar:
@willrogers One widget can have only one class at a time. widgetClass is a "formula-like" property and can be changed any time a certain condition is met (see here https://github.com/ControlSystemStudio/cs-studio/pull/1671#issue-141594283). Autocomplete plugin has already been implemented.

jbobnar:
I have implemented the cascading stylesheets like it is described above. There is also a widgetClass.css inside the opibuilder plugin, which provides some examples.

In the preferences there is an option to add .css files. You can add as many as you wish and they will be loaded in the order as they are listed, overriding predecessors' classes (if the same class is defined in multiple files). When in OPIEditor, you will also notice another toolbar button, which reloads the schema and style sheets and refreshes all currently open OPIs (the widgets that have a class defined will get the updated looks). It is now also possible to assign as many classes to each widget as you wish. The classes have to be separated by space and will be applied in the order as they are listed. For example, you can have a TextUpdate, which has a class property value "yellow largeFont". If the classes .yellow and .largeFont are defined in your css, they will both be applied to the widget.

Opibuilder now depends on two additional plugins net.sf.cssbox.jstyleparser and org.unbescape. In addition you need to have the org.antlr.runtime [3.5.2,4.0.0]. All these plugins are available in maven-osgi-bundles. I've added the two new plugins to the opibuilder feature, while org.antlr.runtime is provided by others, and opibuilder only enforces the version number.

kasemir:
For what it's worth, this is the class support that's now in the display builder.

When you create a *.bcf file (bob class file), the editor changes slightly.
Each property of a widget has a check box next to it to mark this property as being included in the class definition.
For example, drag a Label widget from the palette into the editor.
Set its name to TITLE.
Then check the x, y, font, color and transparent properties
and set them as shown in the screenshot.

class_def

This defines a "TITLE" class for labels: Its position is always in the upper left corner of the display (x and y set to 0). It uses a certain font and colors.

There's a preference setting for listing one or more class files.

When now editing a normal display file, the "class" property of each widget  has a drop-down list that includes "TITLE" as a known class name.
Selecting "TITLE" will apply the x, y, font, color and transparent settings that were defined in the class. Furthermore, these properties are marked with a blue "C" in the property panel, they cannot be modified.

class_use

The display is saved with all the settings of the widget, including those that were provided by the class.
Whenever a display is executed in the display builder runtime, the current class settings are applied.
If the current class files don't define a "TITLE" class for labels, the display will appear as it was last saved. If the current class files do define a "TITLE" class for labels, those will be applied, meaning you can update the class definition files and those new settings will then be applied when a display is executed.

claudio-rosati:
I like it @kasemir.
There only few things I've not understood:

1. When in editor mode of a `.bob` file, how can I define the `.bcf` file to use for my OPI I'm editing?
2. When in runtime, how can I specify the `.bcf` file to use for my OPI(s)?
3. Is it possible/desirable to "embed" a `.bcf` into a `.bob` in order to grant the OPI will always displayed in a certain mode (I think this useful mostly for embedded display: if I'm a company producing a motor, for example, I'll provide the embedded display for controlling the motor and I don't want a _foreign_ `.bcf` breaking my design moving/resizing widgets).

kasemir:
There is a preference setting for a list of class files. (Likewise, the color and font file preference has been turned into a _list_ of color resp. font files.)

So there can be for example a global class file plus a beamline-specific class file.
But at this time there's not yet another class file just for a specific display.
All displays of one CSS instance use the same list of class files. 

The whole point is that one can define TITLE or WARNING_LED classes, and then later change their definition. Displays will use the latest definition.
If don't want your widgets to be affected by class settings, simply don't assign a class to them.

At this time, DEFAULT is always the initial default class for newly created widgets. One could define settings for DEFAULT in the class file. So to be immune to somebody defining DEFAULT, you'd set the class of you widget to an empty string.

For the motor example, if you prefer to use classes, you could use class names like MOT_TITLE, MOT_WARNING, MOT_.. with associated class file.
Whoever uses them can do so without the class file: As long as MOT_* is not re-defined, it'll look as you created the display because the *.bob file has the complete information, it "works" just fine without the class file (just like named colors or fonts work just fine without the color/font file). If users _do_ want to change something, they can look at your class file and add the redefinitions for things they like to change into their site-specific class file.

claudio-rosati:
Understood. Thank you.
A last thing: a `.bcf` file is used to define a single class (like TITLE label), or can be used to define a set of them?

kasemir:
The *.bcf file can have many widgets and many classes.
Note that classes are per-widget. You can have a 'WARNING' class for a Label (where it sets the font and text color), a WARNING class for an LED (where it sets the 'on' color), a WARNING class for a TextUpdate etc. all in the same file.
When you define the same class for the same widget multiple times, either in the same file or in different files, there's a logged warning and the last one wins.

berryma4:
Can the class property be changed by a script?  Or can it be a pv?

berryma4:
First, I should have remarked, thank you very much for this work Kay!

kasemir:
Right now the class cannot be changed at runtime, but that could be added.
So then a script or rule could update the class at runtime.

You can already configure the rules or scripts based on a class.
So a SILLY label class could add a rule to the label that changes the color based on sim://ramp.

kasemir:
The class can now be changed at runtime.
For example, a rule or script could change the class for an LED between ON_OFF and WARNING.
Based on the class, the LED would be colored dark green/green or dark red/orange for off/on..
When a class is applied, the properties defined in the class are set to the class values.
If a property is not listed in a class, it's left alone.

This does mean that toggling between WARNING and DEFAULT might not be useful:
WARNING would set the LED colors to dark red/orange. DEFAULT sets the color - well, not at all, it leaves the color at dark red/orange. That's a side effect of saving the values in the display file, as opposed to a html that would be naked without its style sheet.

berryma4:
Closing because functionality has been added to the display builder (and conversions might be difficult from BOY).

PR #1676, opened 2016-03-18
Clean up of project poms - Issue 7

shroffk:
Associated pull requests are:

https://github.com/ControlSystemStudio/maven-osgi-bundles/pull/18
https://github.com/ControlSystemStudio/cs-studio-thirdparty/pull/16

The primary goal is to standardize all the poms to an extent and also keep them small and manageable

The details are discussed here
https://github.com/ControlSystemStudio/cs-studio-thirdparty/issues/7

Since I am not sure who all are using the profiles being removed these pull requests should be merged after people can verify this change
 - [x] Diamond
 - [x] FRIB
 - [x] ORNL
 - [x] ITER
 - [x] NSLSII
 - [ ] ISIS (not the terrorist group)
 - [ ] ESS

@berryma4 @utzeln @kasemir @willrogers

### DIFF
--- a/applications/alarm/alarm-repository/pom.xml
+++ b/applications/alarm/alarm-repository/pom.xml
@@ -73,55 +73,6 @@
   </build>
   <profiles>
     <profile>
-      <id>csstudio-local-repo-mirror</id>
-      <activation>
-        <property>
-          <name>csstudio.local.repo</name>
-        </property>
-      </activation>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.eclipse.tycho.extras</groupId>
-            <artifactId>tycho-p2-extras-plugin</artifactId>
-            <version>${tycho.version}</version>
-            <executions>
-              <execution>
-                <id>mirror-build-to-local-repository</id>
-                <phase>package</phase>
-                <goals>
-                  <goal>mirror</goal>
-                </goals>
-                <configuration>
-                  <!-- Details: https://www.eclipse.org/tycho/sitedocs-extras/tycho-p2-extras-plugin/mirror-mojo.html -->
-                  <source>
-                    <repository>
-                      <url>${project.baseUri}/target/repository</url>
-                      <layout>p2</layout>
-                    </repository>
-                  </source>
-                  <append>true</append>
-                  <compress>true</compress>
-                  <mirrorMetadataOnly>false</mirrorMetadataOnly>
-                  <destination>${csstudio.local.repo}</destination>
-                </configuration>
-              </execution>
-            </executions>
-          </plugin>
-          <plugin>
-            <groupId>org.eclipse.tycho</groupId>
-            <artifactId>tycho-p2-repository-plugin</artifactId>
-            <version>${tycho.version}</version>
-            <configuration>
-              <finalName>cs-studio-applications-${project.parent.artifactId}-${project.version}</finalName>
-              <repositoryName>cs-studio-applications-${project.parent.artifactId}</repositoryName>
-              <includeAllDependencies>true</includeAllDependencies>
-            </configuration>
-          </plugin>
-        </plugins>
-      </build>
-    </profile>
-    <profile>
       <id>uploadRepo</id>
       <properties>
         <!-- Properties relative to the distant host where to upload the repo -->
--- a/applications/appunorganized/appunorganized-repository/pom.xml
+++ b/applications/appunorganized/appunorganized-repository/pom.xml
@@ -73,55 +73,6 @@
   </build>
   <profiles>
     <profile>
-      <id>csstudio-local-repo-mirror</id>
-      <activation>
-        <property>
-          <name>csstudio.local.repo</name>
-        </property>
-      </activation>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.eclipse.tycho.extras</groupId>
-            <artifactId>tycho-p2-extras-plugin</artifactId>
-            <version>${tycho.version}</version>
-            <executions>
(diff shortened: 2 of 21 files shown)


jbobnar:
I can't tick the checkbox, but this should be **OK** for **ITER**

willrogers:
This might be a problem for us.  Our nightly builds use `csstudio.local.repo`.  I'm all for cleaning up, so perhaps you can help us.

We use this in Jenkins to collect dependencies and pass them between our Jenkins jobs.  Composite repos don't work well for this because the repos are in with the source code so we'd have to zip up the entire repos and pass them around.

The other option I can think of is copying all the p2 repos out after each build and constructing a composite repo from them.  This would be a bit fiddly because we'd end up with maybe 15 components to manage.

willrogers:
Alternatively, we could piggyback `-PuploadRepo` to push the composite repository to a local location.

The only problem here is that the properties aren't quite set up correctly - we'd need a property for the root repository location that we could override to `file:/tmp/repo` or some such thing.

Aside: why is there both a `repository` and a `centralrepository`?

shroffk:
When making changes to a p2 repository it is best to make a copy and perform the changes on this copy. 
In our case, we copy the remote released cs-studio repository to centralrepository and then append the built repository to it.

berryma4:
> Aside: why is there both a repository and a centralrepository?

This is from the git timestamp provider plugin we used in the past to fill the snapshot date with the last git timestamp.  We would diff the central repo with what was built, then only push new plugins.  This was the only way to update the index correctly without pushing the whole thing.

Currently we are using it just for the feature of being able to append the repo (update the index to include past and new plugins).  This is currently necessary for our stable branch retention policy.
(we keep stable branches forever, while mater can be blown away in 24h ... although we hold it longer).

berryma4:
@willrogers currently, I do this ... although I don't like it.  At the end of each build:

```
tar czvf artifacts.tar.gz -C core/base/base-repository/target/repository/ .
```

So, the p2 tar.gz is the artifact for each build.  Then, before a build starts I:
```
mkdir -p "${WORKSPACE}/p2/maven-osgi-bundles"
mkdir -p "${WORKSPACE}/p2/cs-studio-thirdparty"
mkdir -p "${WORKSPACE}/p2/core-base"
mkdir -p "${WORKSPACE}/p2/core-platform"
mkdir -p "${WORKSPACE}/p2/core-repository"
mkdir -p "${WORKSPACE}/p2/core-platform"
mkdir -p "${WORKSPACE}/p2/core-ui"
mkdir -p "${WORKSPACE}/p2/core-utility"
mkdir -p "${WORKSPACE}/p2/core-unorganized"
```
```
echo "<?xml version='1.0' encoding='UTF-8'?>
<?compositeArtifactRepository version='1.0.0'?>
<repository name='CS Studio Composite Repository' type='org.eclipse.equinox.internal.p2.artifact.repository.CompositeArtifactRepository' version='1.0.0'>
  <properties size='3'>
	  <property name='p2.compressed' value='true'/>
	  <property name='p2.atomic.composite.loading' value='false'/>
    <!-- get new time w/ `date +%s000` -->
    <property name='p2.timestamp' value='1395756034000'/>
  </properties>
  <children size='9'>
	  <child location='maven-osgi-bundles'/>
	  <child location='cs-studio-thirdparty'/>
          <child location='core-base'/>
          <child location='core-platform'/>
          <child location='core-repository'/>
          <child location='core-platform'/>
          <child location='core-ui'/>
          <child location='core-utility'/>
          <child location='core-unorganized'/>
  </children>
</repository>" > p2/compositeArtifacts.xml

echo "<?xml version='1.0' encoding='UTF-8'?>
<?compositeMetadataRepository version='1.0.0'?>
<repository name='CS-Studio Composite Repository'
  type='org.eclipse.equinox.internal.p2.metadata.repository.CompositeMetadataRepository'
  version='1.0.0'>
  <properties size='3'>
    <property name='p2.atomic.composite.loading' value='false'/>
    <property name='p2.compressed' value='true'/>
    <!-- get new time w/ `date +%s000` -->
    <property name='p2.timestamp' value='1395756034000'/>
  </properties>
  <children size='9'>
	  <child location='maven-osgi-bundles'/>
	  <child location='cs-studio-thirdparty'/>
          <child location='core-base'/>
          <child location='core-platform'/>
          <child location='core-repository'/>
          <child location='core-platform'/>
          <child location='core-ui'/>
          <child location='core-utility'/>
          <child location='core-unorganized'/>
  </children>
</repository>" > p2/compositeContent.xml

echo "version = 1
metadata.repository.factory.order=compositeContent.xml,\!
artifact.repository.factory.order=compositeArtifacts.xml,\!" > p2/p2.index
```

```
tar zxvf p2/maven-osgi-bundles/artifacts.tar.gz -Cp2/maven-osgi-bundles
tar zxvf p2/cs-studio-thirdparty/artifacts.tar.gz -Cp2/cs-studio-thirdparty
```

I was thinking of making the -PuploadRepo send to our local Nexus machine instead .... 
Which would be the easiest way to lock a version down in the lab.

willrogers:
If you make the root of the `-PuploadRepo` a property, you could change the destination (and so could we).

berryma4:
@willrogers that sounds good.  Although, you may need to add ssh wagon or similar (I believe I only have the s3 wagon)

berryma4:
#1726

berryma4:
@willrogers can I merge this since you will use #1726?

willrogers:
Yes, I'm 90% certain we can use what's in #1726 so I think you can merge this now.

Issue #1678, opened 2016-03-18
VTable widget broken


shroffk:
The VTable widget from the opiwidget.extra is broken due to the inability to resolve its dependencies.
It needs the org.datasource.expressions cannot be resolved.

berryma4:
Thank you!

PR #1681, opened 2016-03-19
#1678 including the expressions package to imported packages

shroffk:
It might be useful to have this patch on the 4.2 branch too


berryma4:
#1678

Issue #1682, opened 2016-03-21
Minor OPIBuilder issues


jbobnar:
The following issues related to opibuilder were found:

* Setting the horizontal alignment has no effect on the multiline text widget
* When linking container is reloaded all its connectors should be marked as "loadedFromLinkedOPI". If not, they are stored twice: once in the child OPI and once in the OPI containing the linking container.
* AbstractSWTWidgetFigure sometimes throws widget disposed exceptions.
* ShowMarcoAction should first display the message dialog and then print to console. Otherwise the console print might shift the focus from a modal dialog to the main window, blocking entire CSS (this is an issue on some linux systems).
* OPIRuntimePreferences: the major/minor severities text for the pulsing property is mixed up
* WidgetService should not allow widgets with duplicate ids.
* Focus traversal with tab and shift+tab should be reversed on the native widgets. Pressing the tab should move the cursor to the next widget, which is the widget that appears later in the opi file. At the moment it goes to the previous, which is non intuitive and inconsistent with the "Next Focus" property. The most obvious is when using the Array Widget.
* Unless specifically needed the dependencies should match the Mars version of eclipse plugins (ITER's build constraint). There is a case of ..e4.ui.workbench in opibuilder and several other dependencies in the new perspectives plugin.

berryma4:
Focus traversal with tab relates to #456

PR #1683, opened 2016-03-21
OPI builder bugfixes

jbobnar:
The pull request resolves the issues listed in #1682.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/META-INF/MANIFEST.MF
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/META-INF/MANIFEST.MF
@@ -35,7 +35,7 @@ Require-Bundle: org.eclipse.ui;resolution:=optional,
  org.python.jython;bundle-version="2.5.2",
  org.csstudio.swt.widgets;bundle-version="2.1.0",
  org.csstudio.swt.xygraph;bundle-version="2.0.4",
- org.eclipse.e4.ui.workbench;bundle-version="1.3.1"
+ org.eclipse.e4.ui.workbench;bundle-version="1.3.0"
 Eclipse-RegisterBuddy: org.python.jython
 Bundle-ActivationPolicy: lazy
 Export-Package: org.csstudio.opibuilder,
--- a/core/ui/ui-plugins/org.csstudio.perspectives/META-INF/MANIFEST.MF
+++ b/core/ui/ui-plugins/org.csstudio.perspectives/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 0.0.1.qualifier
 Eclipse-ExtensibleAPI: true
 Require-Bundle: org.eclipse.osgi.services;bundle-version="3.5.0",
  org.eclipse.e4.core.services;bundle-version="2.0.0",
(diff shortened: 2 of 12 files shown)


berryma4:
@willrogers @shroffk 
I need another reviewer.

willrogers:
Sorry, it's tricky to review because it's a wide range of things.

One thing I would say - I expect that if you build against Mars (not Mars M2) then perspective saving probably won't work correctly now that the versions have been lowered.  It's a bit difficult to test exactly what will happen.

If you build against Mars M2 you should get the latest version of plugins anyway so everything should work ok.

applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/META-INF/MANIFEST.MF
```diff
@@ -35,7 +35,7 @@ Require-Bundle: org.eclipse.ui;resolution:=optional,
  org.python.jython;bundle-version="2.5.2",
  org.csstudio.swt.widgets;bundle-version="2.1.0",
  org.csstudio.swt.xygraph;bundle-version="2.0.4",
- org.eclipse.e4.ui.workbench;bundle-version="1.3.1"
+ org.eclipse.e4.ui.workbench;bundle-version="1.3.0"
```
berryma4: here

core/ui/ui-plugins/org.csstudio.perspectives/META-INF/MANIFEST.MF
```diff
@@ -6,19 +6,18 @@ Bundle-Version: 0.0.1.qualifier
 Eclipse-ExtensibleAPI: true
 Require-Bundle: org.eclipse.osgi.services;bundle-version="3.5.0",
  org.eclipse.e4.core.services;bundle-version="2.0.0",
- org.eclipse.e4.ui.workbench;bundle-version="1.3.1",
+ org.eclipse.e4.ui.workbench;bundle-version="1.3.0",
```
berryma4: and here

berryma4:
@willrogers are these the problem files for the perspective saving?
We need the perspective saving ....

jbobnar:
Tycho will always take the latest from the p2 repository, so you should be safe. The plugin is compatible with both versions.

1.3.1 has one bugfix with respect to 1.3.0. I don't know if this bugfix is relevant or not. There might be additional bugs fixed in future releases, but probably nobody will go back to this plugin and raise this number to 1.4.0. You will just build the CSS using the newest version.

Anyway, I'll ask ITER if they can upgrade eclipse version.

willrogers:
Yes, I think exactly that bugfix, if you look at this: https://git.eclipse.org/r/#/c/63125/

As Jaka says, I think we are safe.  I don't think ITER will get the perspective saving working correctly.  Jaka, how are you building against the older version?  The `core` and `applications` pom files pick up the newer versions.

jbobnar:
Their build server does not have access to internet at all. All eclipse and other plugins were are packed into a p2 repository in advance. They control which versions of plugins are packed and right now its is Mars 0. The generated p2 is installed locally and that is the one that jenkins sees. In addition they also have their own artifactory server, which is a link between the build server and central maven repository, so that they can build things like maven-osgi-bundles.

Issue #1348, opened 2015-09-23
Macros not consistently passed through to child OPIs


nickbattam:
If I define a click-action on a widget to open a new OPI panel with the default 'replace current' mode, the new panel is opened with the parent macros. Any additional macros I have specified on the widget are not passed though and the 'pass parent macros' choice is ignored.

If I change the 'mode' to open in a new standalone window or new tab the macro definitions I specify are correctly passed though to the new panel.

I have been able to reproduce this behaviour with a number of different widgets.

This problem is seen running the Diamond fork of the 'master' branch. Our fork contains no changes that should affect this behaviour.

PR #1693, opened 2016-03-29
Override child OPI macros

nickbattam:
This resolves issues seen in #1348 by ensuring any macros passed into an OPI when it is opened replace the default values


berryma4:
If it's okay, we will be having a code a thon next week.  I would like to discuss macros in general. #1600

nickbattam:
@berryma4 That's fine

berryma4:
Please merge master too this branch.  A Travis ci yml file was added for testing.

willrogers:
So perhaps we can revisit this now.

* Default macros on a screen (1) are lowest priority
* Macros attached to a button or a linking container on that screen (2) override (1)
* Macros passed in to the screen at launch (3) override both: the screen AND any linking containers AND any screens launched from buttons

Is my logic incomplete?

kasemir:
I would prefer a priority that is similar to shell variable definition/overrides, or any variable overrides in nested scenarios for that matter:

1) Preferences may define macros
2) A button or linking container that opens a display can add/replace macros
3) The display file itself can add/replace macros
4) A grouping container inside a display can add/replace macros for its child widgets

That is pretty easy to understand because the layering is as in every other variable-based scenario.

The difference to BOY and your proposal is the handling of macros in the display file itself. Should they replace what's passed to the display? I think so, just as you'd have in pretty much any programming language:
````
run_some_display(x = "Value passed in"):
   x = "My value"
  print(x)
  # --> will print "My value"
````

BOY, however, behaves more like this:
````
run_some_display(x = "Value passed in"):
   # Only used in case 'x' has no value
   x = "My value"
  print(x)
  # --> will print "Value passed in"
````

Doesn't make any sense to me. OK, I get that this may be convenient to set a default for testing a screen, but it still doesn't make much sense because it's hard to remember how the variables are overwritten. So for testing, define a top-level test screen that provides macro values for the screens you want to test. Just like you need to provide parameters to functions when you want to unit-test a function.

willrogers:
OK, that's quite clear and I might just have been wrong all along.

My confusion arose from the fact that it is possible to attach macros to a screen.  These _aren't_ defaults - they just win.  ~~These seem to have no value whatsoever.~~

Actually there's one possible use case.  You can set macros as defaults on the very top-level screen and have the ability to override them lower down.  You could do that with global preferences, but this way you could override the global preferences?  That seems like a fairly niche case.

The particular problem that we were trying to solve can be solved using a global macro definition - I didn't know these existed.

Thanks for explaining all that again.

Issue #1696, opened 2016-03-30
PGCopyPreparedStatement hardcoded name issue


jbobnar:
..archive.writer.rdb/src/org/csstudio/archive/writer/rdb/PGCopyPreparedStatement.java is using a hardcoded table name "sample".

PR #1697, opened 2016-03-30
Archive writer hardcoded table name fix

jbobnar:
Fix of the issue #1696.

In addition, javadoc syntax has been fixed in another archive related class.


berryma4:
@kasemir okay?

kasemir:
Fine, all about Postgres and diirt.

Issue #1705, opened 2016-04-04
remove products directory


berryma4:

willrogers:
No more products.

PR #1706, opened 2016-04-04
remove products directory

berryma4:
#1705


shroffk:
The products being deleted include KEK and panda.
There is a logbook factory extension point, in o.c.logbook.midas which is being removed. It should be either added to a site specific repository or to the application/logbook

Issue #1729, opened 2016-04-07
Data Browser: Allow configuration of 'Area's opacity


kasemir:
The change from the 'average' line and the 'min/max' area color is currently fixed:
Line uses the color, area uses a more translucent version of that color.

--> Make the amount of translucency configurable.

berryma4:
Suggestion taken from  BNL Operations feedback presentation.

berryma4:
This is great!!!  Thank you so much @kasemir !

berryma4:
@tashasummers @shroffk

PR #1747, opened 2016-04-12
Data Browser: Add 'opacity'; font fixes

kasemir:
Adds 'opacity' to allow configuring the area as requested in #1729

Noted that the label and scale fonts weren't actually used, so enabled them.


berryma4:
The first part only fails because the branch was made before the travis yml file was added.  merging.

Issue #1770, opened 2016-04-26
AnnunciationFormatter fails with the new format rules


shroffk:
At NSLS2 we use `{}` in our PV names, the new forced format option in cause parsing errors.

e.g. the Alarm description includes the pv name 
XF:31IDA-OP{Tbl-Ax:X1}Mtr

When this description goes through the new AnnunciationFormatter with the forced formatting, it causes a flurry of illegal argument exceptions since the formatter is trying to resolve the Tbl-Ax:X2 as a special custom format parameter in XF:31IDA-OP**{Tbl-Ax:X1}**Mtr

`
java.lang.IllegalArgumentException
	at org.eclipse.osgi.util.NLS.internalBind(NLS.java:192)
Caused by: java.lang.NumberFormatException: For input string: "Tbl-Ax:X2"
`

This was added due to 
Bugzilla ticket 7491: Alarm Table - alarm description without alarm severity prefix;

Potential solutions:
Could it be possible to make the forced formatting a preference?
or
Sanitize the format string before giving it to the NLS.bind

@kasemir @utzeln @jbobnar

kasemir:
Do you have a link to the commit or pull request for the new AnnunciationFormatter?

jbobnar:
@kasemir 
https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/alarm/alarm-plugins/org.csstudio.alarm.beast/src/org/csstudio/alarm/beast/AnnunciationFormatter.java

@shroffk 
The commit you mention was related to this #1281. I'm sure this could be made a preference. I'll make a pull request with the change.

shroffk:
Thanks

kasemir:
Bummer.

The use of '{0}', '{1}', .. in the annunciation strings has been in use for a while.
PVManager decided to use '{ .. }' for ill-advised modifiers like 'longString'; still, it's using curly brackets as well. Putting such brackets into the PV name was an unfortunate idea.

Still, too late now.

How about replacing the NLS.bind() calls with more specific String.replace() calls that replace only "{0}", "{1}", "{2}" with the respective replacement text?

shroffk:
> How about replacing the NLS.bind() calls with more specific String.replace() calls that replace only "{0}", "{1}", "{2}" with the respective replacement text?

I think this would be a good idea, to sanitize the input string before giving it to bind

```
format = format.replace("{", "'{'"); 
format = format.replace("}", "'}'");  
[...]

PR #1772, opened 2016-04-27
o.c.alarm.beast.test: Re-enable tests

kasemir:
Was found disabled in #1770.
Needed to be updated to Instant and new location of alarm plugins.


kasemir:
Kunal just did the same thing and then some..

PR #1781, opened 2016-05-05
o.c.java: Update test

### LINKED ISSUE #1779: scan data display time stamp for later devices
The scan data log contains the 'serial' of the log command invocation and the original time stamp for each logged sample.

In the table display, values are aligned by 'serial'. The individual time stamps are shown in the tool tip of each cell, and an overall time stamp is computed as the max. time of all samples in the row.

The overall time stamp for the row is misleading for this case:

Serial | PV 1      | PV 2      | PV 3
-------- | ---------- | ---------- | --------------
   0      | Value 1 | Value 2 | -nothing-
   1      | Value 3 | Value 4 | Value 5

A bug uses the time stamp of Value 5 as the overall time stamp for the serial 0 row, because it already found the initial value for PV 3, even though PV 3 doesn't have anything for serial 0.

kasemir:
Recent pull request  #1780 to fix #1779 requires this adjustment of test because Time stamp format (date part) has changed.

The pull request added about 20 tests from the o.c.scan.test plugin, but regrettably I missed this test in o.c.java. Shows that it's good to have the tests run automatically!

### DIFF
--- a/core/platform/platform-plugins/org.csstudio.java.test/src/org/csstudio/java/time/TimestampFormatsTest.java
+++ b/core/platform/platform-plugins/org.csstudio.java.test/src/org/csstudio/java/time/TimestampFormatsTest.java
@@ -23,12 +23,12 @@ public class TimestampFormatsTest
     @Test
     public void testFormatter()
     {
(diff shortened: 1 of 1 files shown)


berryma4:
It seems Travis only fails on build fails, I might need to change a property to return 1 on failed tests.

berryma4:
#1782

PR #1789, opened 2016-05-16
Save set restore

berryma4:
As discussed in the last meeting, merging save set restore into an applications directory.
@shroffk 
#1516


jbobnar:
The root cause of the new errors is similar as before. The FX Widget Spinner has only been added in Java 8u40 (https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Spinner.html). It seems that Oracle is doing major FX changes in between java release.

Which java version is used on the build server? I tried 8u31 and it doesn't show any other problem than the spinner; but if anything older is used there might be other issues as well. I'll have a look if I can replace the spinner with another (custom) widget

kasemir:
Try at least 1.8.0_70.
The display builder needs 1.8.0_40 because of javafx.scene.control.Dialog. Similar to the Spinner, new things were added in _40.
In 1.8.0_51 there was a Combo problem on Windows. So overall I'm using at least _70.

berryma4:
I'm seeing about updating the version on Travis

berryma4:
It looks like I can force it to update, and grab 8u91.  @jbobnar, I should probably revert your past two changes.

jbobnar:
No need to revert. It is OK as it is. The TinySpinner class is now more by the book than it was before.

I also pushed another commit today in the morning, fixing one cosmetic bug.

berryma4:
@jbobnar, I still get:
tycho-surefire-plugin return code 13
for:
org.csstudio.saverestore.git.test
org.csstudio.saverestore.masar.test

locally and on travis-ci

jbobnar:
It seems to be a mockito issue. For some reason tycho pulls in mockito, which depends on org.hamcrest. I made a PR on the save-set-restore branch and also updated the FRIB repository.

berryma4:
Thank you @jbobnar and @kasemir !!

PR #1800, opened 2016-06-10
vtype.pv: loc:// support for VTable

kasemir:
Allows declaring local vtype.pv as <VTable>, and/or writing a VTable to local PV.

Part of #1429


kasemir:
VTable support hadn't been merged, yet. Added DBE_PROPERTY support to same pull request since they're both in vtype.pv.

PR #1830, opened 2016-07-20
scan server pvmanager support

berryma4:
#1829


berryma4:
@shroffk thoughts?
I have a few questions, and changes I would like to make.

Issue #1857, opened 2016-08-17
Update vtype.pv to match updated caj and pvaccess libs


kasemir:
Dirt updated the libs, and maven-osgi-bundles packages new ones in org.csstudio.platform.libs.epics.
See
https://github.com/ControlSystemStudio/diirt/commit/804addd5355498df9baef9e3ee8d96c48f585a05
#1855

kasemir:
Additional quirk: pvaccess plugin uses import-package for caj with a specific version. Need to include caj plugin in product, just o.e.platform.libs.epics which wraps jca and caj is not sufficient.

--> Remove the dependency from vtype.pv on o.c.platform.libs.epics. Copy preference handling code code from o.c.platform.libs.epics into vtype.pv, using the existing preference names.

PR #1858, opened 2016-08-17
vtype.pv: Update versions of pvdata and pvaccess

kasemir:
Fixed vtype.pv compilation error after dirt & maven-osgi-bundles updated to new PVA libs
#1857


kasemir:
The version change is currently blocking the build.
This pull request will allow 'core' to build again.

I am planning to submit more changes onto the branch to remove the dependency on o.c.platform.libs.epics. Since vtype.pv seems to be the only plugin that uses it, the relevant code from o.c.platform.libs.epics will be copied into vtype.pv, using the existing preference names because those have been in effect for ~10 years and changing them would be disruptive.

berryma4:
Also, the bundle names are being changed back:
https://github.com/epics-base/pvDataJava/issues/12
https://github.com/epics-base/pvAccessJava/issues/8

PR #1868, opened 2016-08-23
opibuilder: Fix ArrayIndexOOB when writing long strings w/ Format=String (4.3.x)

### LINKED ISSUE #1834: BOY {"longString":true} vs. Format=String, ArrayIndexOutOfBoundsException
For long strings held in character waveforms, text widgets can use either
 * A PVManager PV that ends in {"longString":true} 
 * The Format=String option of the widget

The latter is much easier to remember, and results in a PV name that can be used in other tools.
But there is a problem with the Format=String option for Text Input widgets:
Trying to enter a string that's longer than the current value can result in ArrayIndexOutOfBoundsException, and the value is not written to the PV.

nickbattam:
Fixes #1834 on the 4.3 branch


berryma4:
thank you!

PR #1872, opened 2016-08-24
Fail the build on whitespace violations.

willrogers:
You don't have to merge this - I'm just trying to prove that this will fail the Travis build.

Assuming there are some violations...


willrogers:
Like that ...

willrogers:
Now I'll fix the violations.

willrogers:
See discussion at https://github.com/diirt/diirt/pull/70

willrogers:
Ironic.  Failed because of trailing whitespace in `checkstyle-output.xml`.

berryma4:
thank you!

berryma4:
Kind of annoying, why does tycho surefire return 0 when tests fail?
I added this line because it doesn't, so I can check:
https://github.com/ControlSystemStudio/cs-studio/blob/master/build/travis.sh#L16

willrogers:
Is it just this?

https://github.com/ControlSystemStudio/cs-studio/blob/master/core/pom.xml#L398

berryma4:
hilarious!  I should have grep'ed instead of trusting my eyes :)

Issue #1877, opened 2016-08-25
Update org.csstudio.archive.reader.ea4 to use new pvAccessJava


kasemir:
Another side effect of the V4 changes, #1859:

````
Missing requirement: org.csstudio.archive.reader.ea4 1.0.0.qualifier requires 'bundle org.epics.pvaccess 3.0.0' but it could not be found
````

Needs to use org.epics.pvAccessJava 4.1.3 instead of org.epics.pvaccess 3.0.0.
May also need to update reference to pvdata/pvDataJava

kasemir:
@shroffk , do you have a way to update the pvAccess reference and test this at BNL?

PR #1879, opened 2016-08-25
archive.reader.ea4: Update pvdata, pvaccess references

kasemir:
Fix compilation of archive.reader.ea4 for new E4 libs.
#1877
@shroffk Kunal, do you have a way to test this at BNL with an actual archive data server?


shroffk:
I don't have pvaccess data being archived in production, I am working on setting up an example v4 AA.

kasemir:
In that case, just having it compile might be good enough.
The E4 API hasn't changed. For the vtype.pv, PVA worked after I switched to the new E4 plugins.

shroffk:
For the ea4 library compiling is enough

Issue #1903, opened 2016-09-21
Unnecessary error logging


willrogers:
Here: https://github.com/ControlSystemStudio/cs-studio/blob/534fea1d38e6cbef925548fdf4aa4943290f0a9c/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/src/org/csstudio/opibuilder/runmode/LauncherHelper.java#L60

an error is logged when a file is opened that is not inside the Eclipse workspace:

```
2016-09-21 14:38:22 ERROR: A workspace match for /tmp/widgets.opi could not be found.
null
```

This happens when you open a file using File -> Open.  CSS handles this reasonably well - you can run the file in the workbench, save the file back to the location outside the workspace, and 'Save As...' the file into the workspace.

So I think this message should be downgraded to a warning or info.

PR #1904, opened 2016-09-21
Reduce logging severity when file is outside workspace.

willrogers:
@jbobnar do you know of a different way to get to this error?

See #1903.

Should it be warning rather than info?


jbobnar:
It was related to linked resources. See the discussion here: #1302 
I don't remember what was the actual usecase, but I think info is OK. I think that in case of an error, there are other logs as well. That should be sufficient for the user.

Issue #1914, opened 2016-10-03
OPI shells are not correct size on launch


willrogers:
OPI shells should open so that they are exactly large enough to contain the OPI within.

We previously did not get this correct, but I have got a better mechanism similar to the one in OPIView.

PR #1915, opened 2016-10-03
Automatically resize to contents when opening OPIShell.

willrogers:
See #1914


willrogers:
(This one should not be controversial).

PR #1947, opened 2016-11-07
Check Box widget auto size issue

rosarp:
Fixed #1943

PR #1950, opened 2016-11-07
o.c.swt.rtplot.test: Increase times in time-related test

kasemir:
Should help avoid test failures as seen in #1947


berryma4:
Thank you!

kasemir:
Glad that the adjusted delays make the test pass - for now..

Issue #1946, opened 2016-11-07
Support for read-only PVs in saverestore


jbobnar:
Saverestore should provide an option to define a PV as read-only. Such a PV should be saved when the snapshot is taken, but would not be restored.

berryma4:
Thank you so much!!!

jbobnar:
This is done. Closing.

PR #1952, opened 2016-11-10
Save restore read only

jbobnar:
This is related to #1946.

It is possible to define read only PVs in the save set files. All existing ones (where this property is not defined) will behave as if none of the PVs is read only (they are all writable). 

When you open a snapshot you cannot select/unselect the read-only PV (the checkbox is disabled). However, I included an option to overwrite the read-only property through the popup menu. If that is not desirable, let me know and I will remove it (or make it configurable).

The amount of changes is quite large, because I also redesigned the save set and snapshot classes. It should now be much easier to add any additional properties to the save set.


berryma4:
The test failure is not related to the changes;
```java 
Test set: org.csstudio.swt.rtplot.util.UpdateThrottleTest
-------------------------------------------------------------------------------
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.035 sec <<< FAILURE! - in org.csstudio.swt.rtplot.util.UpdateThrottleTest
testUpdateThrottle(org.csstudio.swt.rtplot.util.UpdateThrottleTest)  Time elapsed: 5.019 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: <2>
     but: was <3>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.csstudio.swt.rtplot.util.UpdateThrottleTest.testUpdateThrottle(UpdateThrottleTest.java:65)

```

Issue #1955, opened 2016-11-14
ScanClient API change causing jython scan examples to fail


rosarp:
There is a change in API under [#6bd0022991](https://github.com/ControlSystemStudio/cs-studio/commit/6bd0022991ee01783d5e2a04fa7564439e34be52#diff-f49fef019fd53c001fda393021f1be79) revision for ScanClient.

Which is causing [scan_client.py](https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/scan/scan-plugins/org.csstudio.scan/jython/scan_client.py#L131) to fail, since submitScan method now needs 3rd argument.

Based on that also scan examples fails. One of such examples is [custom_scan.py](https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/scan/scan-plugins/org.csstudio.scan/examples/custom_scan.py#L11)

Probable solution for scan_client.py is as follows:

```
diff --git a/applications/scan/scan-plugins/org.csstudio.scan/jython/scan_client.py b/applications/scan/scan-plugins/org.csstudio.scan/jython/scan_client.py
index 80547a4..54003dc 100644
--- a/applications/scan/scan-plugins/org.csstudio.scan/jython/scan_client.py
+++ b/applications/scan/scan-plugins/org.csstudio.scan/jython/scan_client.py
@@ -112,7 +112,7 @@
         return self.client.simulateScan(xml)
 
 
-    def submit(self, name, commands):
+    def submit(self, name, commands, queue):
         """
         Submit a CommandSequence to the server for execution
         
@@ -128,7 +128,7 @@
             xml = commands.getXML()
         else:
             raise Exception('Expecting CommandSequence or XML-text')
-        self.id = self.client.submitScan(name, xml)
+        self.id = self.client.submitScan(name, xml, queue)
         return self.id
 
     def getScanInfo(self, id=-1):
@@ -289,6 +289,9 @@
             args.pop(0)
         else:
             name = "Scan"
+        # This is bool value for queue
+        queue = args[0]
+        args.pop(0)
             
         # End result, overall scan
         cmds = self._decodeScan([], args)
@@ -296,7 +299,7 @@
             raise Exception('Empty scan')
 
         seq = CommandSequence(cmds)
-        id = self.submit(name, seq)
+        id = self.submit(name, seq, queue)
         if __name__ == '__main__':
             seq.dump()
             self.waitUntilDone()
```

kasemir:
[...]

PR #1956, opened 2016-11-14
Scan examples 1955

kasemir:
Fixes issue in legacy jython client library, #1955,
then deprecates it and updates the examples to
use PyScanClient examples


kasemir:
Thanks for the fixes. Hadn't noticed that the Eclipse editor "save actions" that cleanup white space are limited to Java sources, not arbitrary text files incl. html.

Do you still see that timing-related test failure?

berryma4:
The timing failure hasn't come up since.  Although, I did have to change the travis file.  They are using a new container, and the jdk line seemed to make it fail.

Issue #1965, opened 2016-11-22
save restore null pointer


berryma4:
@jbobnar do you run into this when starting with a new repo?

```java
Exception in thread "JavaFX Application Thread" java.lang.NullPointerException
 at org.csstudio.saverestore.ui.browser.DefaultBaseLevelBrowser.confirmSelectBaseLevel(DefaultBaseLevelBrowser.java:121)
 at org.csstudio.saverestore.ui.browser.DefaultBaseLevelBrowser.lambda$2(DefaultBaseLevelBrowser.java:86)
 at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
 at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
 at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
 at com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
 at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
 at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
 at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
 at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
 at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
 at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
 at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
 at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
 at com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
 at com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:49)
 at javafx.event.Event.fireEvent(Event.java:198)
 at javafx.scene.Node.fireEvent(Node.java:8413)
 at javafx.scene.control.Button.fire(Button.java:185)
[...]

PR #1967, opened 2016-11-22
Save Restore null pointer fixed

jbobnar:
Bug fix for #1965.

It was a "typo" kind of a bug. Instead of going through a method the field was accessed directly.


berryma4:
thank you!

Issue #1970, opened 2016-11-28
Databrowser improvements


willrogers:
Our operators have just started using databrowser in earnest and have identified a few problems:

* ~~In Time Axis properties you can edit relative times and the plot will update.  If you edit absolute times nothing happens.~~
* ~~If absolute times are selected and you press the ... button, both times currently are set to 2016-02-20!~~
* In Value Axis properties, if you change Min or Max autoscale should be turned off
* ~~If you open Inspect Waveforms, select a PV, and press refresh, you get an error message:~~ (fixed in #1971)
** `2016-11-29 17:33:39.290 WARNING [Thread 1] org.csstudio.swt.rtplot.internal.AxisPart (setValueRange) - Axis Waveform Index: Bad value range 0 ... 0`
** The x-axis range does not rescale until you move the slider

There are also some feature requests:

* Should PV searching be case-insensitive by default?
* A bunch of possible improvements to the 'Inspect Waveforms' view:
  * ~~Autoscale the Y axis by default when opening a waveform~~ (#1975)
  * ~~Exceptions thrown when Inspect Waveforms view is open~~ (#1973)

I'm happy to do some of this work, but before we start I'd like to know the plan for databrowser 3:

* Will it reuse any of the current code so we can fix things for both?
* Will it be released very soon so there's no point fixing problems in databrowser 2?

utzeln:
We notice that when you edit absolute times, the auto scale of the traces are switched off, which gives often the impression that nothing happens. Actually, if you check the auto scale of each trace, then the plot is displayed correctly.
This happens even if in the preferences, auto size setting is checked.

kasemir:
> If absolute times are selected and you press the ... button, both times currently are set to 2016-02-20!

There was a problem with parsing the times and thus showing the wrong start/end time in the dialog because of the time format change. Has been fixed with https://github.com/ControlSystemStudio/cs-studio/pull/1933

kasemir:
The case sensitivity of searches is handled by the archive.reader, outside of the data browser 2/3 code. I'm OK with updating that to case-insensitive, but you'd have to do that separately for channel archiver, rdb (likely different SQL dialects), archive appliance, DLS fast/short term archiver, ...

kasemir:
As for the databrowser 2/3 plan:
Databrowser 3 is currently in the same repository as the display builder.
[...]

PR #1971, opened 2016-11-29
Databrowser: correctly determine size of waveform in view.

willrogers:
Discussed in #1970.

@kasemir it may be worth having a `VTypeHelper` method to find size of any VType.


kasemir:
Thanks for fixing this, and also for having pull requests specific to just one issue.
Looks good to me.

kasemir:
Has been included in data browser update

Issue #1973, opened 2016-11-30
Invalid thread access from Inspect Waveforms view


willrogers:
To reproduce:

* Open databrowser plot
* Right click | Inspect Waveforms
* Add a PV to the databrowser plot

```
2016-11-30 09:38:28.727 WARNING [Thread 74] org.diirt.util.concurrent.Executors$2 (execute) - Exception on the timer thread caused by a ValueListener
org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4491)
	at org.eclipse.swt.SWT.error(SWT.java:4406)
	at org.eclipse.swt.SWT.error(SWT.java:4377)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:482)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:420)
	at org.eclipse.swt.widgets.Combo.getSelectionIndex(Combo.java:1087)
	at org.csstudio.trends.databrowser2.waveformview.WaveformView.update(WaveformView.java:240)
	at org.csstudio.trends.databrowser2.waveformview.WaveformView.access$0(WaveformView.java:221)
	at org.csstudio.trends.databrowser2.waveformview.WaveformView$1.changedItemLook(WaveformView.java:91)
	at org.csstudio.trends.databrowser2.model.Model.fireItemLookChanged(Model.java:937)
	at org.csstudio.trends.databrowser2.model.ModelItem.fireItemLookChanged(ModelItem.java:198)
	at org.csstudio.trends.databrowser2.model.ModelItem.setUnits(ModelItem.java:169)
	at org.csstudio.trends.databrowser2.model.PVItem.updateUnits(PVItem.java:393)
	at org.csstudio.trends.databrowser2.model.PVItem.pvChanged(PVItem.java:380)
	at org.diirt.datasource.PVReaderImpl.firePvValueChanged(PVReaderImpl.java:110)
	at org.diirt.datasource.PVReaderImpl.setValue(PVReaderImpl.java:220)
	at org.diirt.datasource.PVDirector$4.run(PVDirector.java:358)
	at org.diirt.util.concurrent.Executors$2.execute(Executors.java:55)
	at org.diirt.datasource.PVDirector.notifyPv(PVDirector.java:323)
	at org.diirt.datasource.PVDirector.access$500(PVDirector.java:38)
[...]

PR #1979, opened 2016-11-30
Databrowser: check if widget is disposed before updating.

willrogers:
@kasemir see #1973 and #1970.


kasemir:
Thanks!

willrogers:
Another change where if the Inspect Waveform view has a PV selected but it is removed from the model, it needs to be removed as the selected item.

kasemir:
Included in new data browser.

Issue #1980, opened 2016-12-01
Feature request: Waveform view should be labelled on main databrowser plot


willrogers:
We had a feature request from operators:

> When you have the Inspect Waveform view open, can you tell where on the main databrowser plot you're viewing?

I think this is a good idea, but the question is how best to do it?  I have a prototype that draws a label on the time axis as shown in the picture, and this seems to work quite well.  It requires a bunch of changes around having a new member variable `private XTYPE x_axis_label` in `Plot.java`.

![db-label](https://cloud.githubusercontent.com/assets/2440055/20788476/dbcfd20c-b7a8-11e6-89f4-3b4525c6edb0.png)

Thoughts?

See also #1970.

kasemir:
How about updating the X axis title. Instead of "waveform index", it could show the time stamp of that waveform.

kasemir:
Sorry, somewhat misunderstood. You're talking about an indication in the 'source' data browser.

How about adding/moving an annotation? It would better identify which trace and then the position on that trace that's inspected in the waveform view?

willrogers:
This was before your new comment, but for the record:

---

The timestamp is actually in the box at the bottom of the waveform view (that's not new).  So the information is there already.

The change is the little label (06:53) on the x-axis of the main plot (just in case you didn't notice that).  The label moves around as you drag the slider.  It might equally be a little red dot on the axis or something else - I just found that a label was easy to draw.

The visual feedback is nice because it is an easier way to understand the link between the two plots.  It also shows you when the data is more or less frequently sampled, which we've already found to be a bit surprising.

The question (as I see it) is whether all the change to the model and plot is worth-while.

willrogers:
An annotation sounds like it might be a good idea, plus it would be less invasive to the existing model.  Let me try that.

berryma4:
Should I  keep #1981 open for you?

kasemir:
Right, the annotation support is already "there", so hopefully this is easier to implement:
Waveform view updates the annotation, and plot already automatically reflects that.

On the other hand, the user can currently move an annotation.
Should this waveform-created annotation be locked from user interaction?
Or should moving the annotation actually select a different instance of the waveform??

willrogers:
@berryma4 just for now while we discuss things please.

PR #1981, opened 2016-12-01
Databrowser: show waveform view timestamp on main plot.

willrogers:
This requires:

* x_axis_label variable in Model and Plot classes.
* changes to interfaces and adapters for changing this
* changing WaveformView to set this variable in the model

See #1980 and #1970.

This is to show what was necessary for the prototype, and is not necessarily for merging.


berryma4:
@willrogers did you have another commit to add to this?

willrogers:
I've been looking at this but it's tricky, so nothing yet.

willrogers:
Actually I'll close this and we can discuss the other changes separately.

Issue #1975, opened 2016-11-30
Inspect Waveforms view should autoscale Y axis by default


willrogers:
The Inspect Waveforms view doesn't autoscale the Y axis.  This makes it a bit fiddly to use unless the values happen to be in the range 0-10.

The recent changes to autoscale mean that autoscale can be turned on by default and any further manual changes will turn it off, which seems like better behaviour.

kasemir:
Thanks for fixing this!

PR #1984, opened 2016-12-06
1975 waveform scale

willrogers:
See #1970 and #1975.  Some minor improvements.


kasemir:
Thanks, I've included this in the data browser update

Issue #1986, opened 2016-12-07
EDM Converter: More accurate conversions of widget sizes


mfurseman:
All the widgets in the converter use the same logic for determining x and y placement, and for selecting the width and height. This doesn't do a good job of converting certain EDM widgets that draw their border outside of the specified container, something CS-Studio is not able to do.

PR #1987, opened 2016-12-07
EDM Converter: fine tune locations and sizes of widgets to better match EDM

mfurseman:
This pull request address issue #1986

This branch adds the ability to override tags in the XML, then uses that functionality to configure placement of each individual type of widget.

Previously, if a tag name was reused the tag would appear twice in the resulting XML, the change to OpiAttribute removes the old tags before creating a new one. This allows widgets which require different conversion logic to implement their own x, y, width and height conversion logic after the superclass has already set these values. It also tidies up the XML where a number of these values were already being overridden.


mfurseman:
After fixing the failing tests I am now happy with the state of this branch.

Issue #1990, opened 2016-12-08
Inconsistent text size on TextUpdate widget


mfurseman:
We have an issue where the text update widget has inconsistently sized text each time the OPI is refreshed. We've tracked this down to code written to solve https://github.com/ControlSystemStudio/cs-studio/issues/107.

You can reproduce this by creating a text update widget with a small height, and attaching a local enum PV. I've attached the [OPI](https://github.com/ControlSystemStudio/cs-studio/files/639554/TextUpdate.opi.txt)  from the screenshots below.

![screenshot-working](https://cloud.githubusercontent.com/assets/5279818/21011827/ff27f386-bd4a-11e6-9929-70a07422faf2.png)
![screenshot-not-working](https://cloud.githubusercontent.com/assets/5279818/21011833/0993e2da-bd4b-11e6-864b-d9890c4e6601.png)

What appears to be happening is that when the OPI is originally drawn the PV is disconnected and the disconnected overlay is drawn at the top of the widget. The code designed to fit text into the widget then tried to resize it to fit into the widget which no longer has a finite size, and only stops reducing the size of the text once the PV is connected. Removing the section of code that does that resizing fixes this issue for us, however I assume that's a breaking change for people who were affected by  https://github.com/ControlSystemStudio/cs-studio/issues/107. Any good ideas for a solution to this?

berryma4:
I feel the code written to solve #107 was a pretty bad patch to the larger issue of fonts and screen resolution.  I think @shroffk was more correct with his comment:
https://github.com/ControlSystemStudio/cs-studio/issues/107#issuecomment-22948334

So, it would probably be best to remove the #107 code, and replace it with @shroffk comment.

kasemir:
Generally agree that font sizes in the display file should be in pixels, just like all the other sizes and locations in the display file. The display builder uses that approach. Somewhat too late for BOY. Auto-resizing widgets to fit the contained text becomes less important/useful when the font size you see at design time won't change on you, because it's kept in pixels.

So I think I'm OK with removing the resize "fix", but I'm not 100% sure what code we're referring to, so please post a pull request so we see the exact change.

mfurseman:
This pull request just removes relevant code, it doesn't attempt to implement @shroffk code yet.

kasemir:
Changing BOY to use pixels instead of 'points' for font sizes would be a big change, not worth it.

Just removing that font-shrinking code may also cause quite some upset, as Eric commented.

[...]

PR #1991, opened 2016-12-08
Remove font resizing code from issue-107

mfurseman:
Do not merge yet. This is for discussion on https://github.com/ControlSystemStudio/cs-studio/issues/1990.


berryma4:
So, removing this will mess up quite a few peoples screens.
But, replacing it with a pixel correction, I think everyone would be happier.

berryma4:
Do you want to close this, and open a new pull request with what was discussed on #1990?

willrogers:
Yep we'll do that.

PR #1993, opened 2016-12-08
First attempt at labelling waveforms with annotations.

willrogers:
This isn't for merging - I need some help from @kasemir.  See #1980 and #1970.

There are a number of reasons why this is tricky:

* At present, annotation changes are passed only from the plot back to the model, not from the model through to the plot
* The annotation is very happy to draw itself outside of the plot window
* The annotation is saved into the `.plt` file, but it shouldn't really be because it's only associated with the optional waveform view
* There are two lists of annotations - `AnnotationInfos` on the model side and `Annotations` on the plot side.  These need to be kept in sync and since they are both immutable it's a bit of work

This commit solves some of these problems, but not necessarily in a good way:

* The waveform plot annotation is identified simply by having the name "Waveform view"
* The best behaviour would be for the annotation to be draggable and the plot to be updated.  This works but is so slow that you can drag the annotation at most one point

Perhaps Kay can see some simpler ways to do things.

### DIFF
--- a/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/model/AnnotationInfo.java
+++ b/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/model/AnnotationInfo.java
@@ -84,6 +84,23 @@ public class AnnotationInfo
         return "Annotation for item " + item_index + ": '" + text + "' @ " + TimeHelper.format(time) + ", " + value;
     }
 
+    @Override
+    public boolean equals(Object other)
+    {
+        boolean value = false;
+        if (!(other instanceof AnnotationInfo))
+            value = false;
+        if (other == this)
+            value = true;
+        AnnotationInfo otherAnn = (AnnotationInfo) other;
+        value = this.item_index == otherAnn.getItemIndex() &&
+                this.time.equals(otherAnn.getTime()) &&
+                this.value == otherAnn.getValue() &&
+                this.offset.equals(otherAnn.getOffset()) &&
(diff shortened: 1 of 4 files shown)


kasemir:
Thanks for trying this.
I think you found all the quirks.
Still, do you agree that the end result would be the right thing? I.e. that is a good way to indicate the currently selected waveform sample?
Then I'll pull that into a branch and see if I come up with some adjustments, but overall I think you're on the right track.

willrogers:
If we ignore the difficulties, I think this is a very useful feature and an annotation is good way to indicate it.  Some ideas:

* Allow an annotation to be 'not saveable'.  This solves some problems.  However, when the two lists of annotations are immutable, if someone changes an annotation on the plot side, how do we know that it is the 'Waveform view' annotation that has changed?
* If an annotation is off the main plot window, draw a little arrow of the correct colour at the edge of the plot pointing at the location of the annotation.

The main remaining problem is getting the waveform plot to update when triggered by `changedAnnotations()`, here: https://github.com/ControlSystemStudio/cs-studio/pull/1993/files#diff-6e378a0d471757ce4cb7f514f7e18ab3R126.  Pushing back on to the UI thread with `asyncExec()` or `syncExec()` disrupts the mouse move in some way.

kasemir:
Right, the annotations a flag like 'transient', not meant to be saved. Maybe also not meant to be moved by the user, if that part becomes too hard because of slow-down, jerky movements etc.

I'll look at this by early next week.

kasemir:
I've copied the code changes into a branch 1980_waveform_annotation to add the 'transient' annotation handling, add flags to prevent loops when waveform view updates annotations in model and then sees those changes from model etc.
Closing this pull request because we'll then have one for the branch in this repo.

Issue #1988, opened 2016-12-07
Intermittent deadlock saving databrowser plot to file on Linux


willrogers:
On Linux:

* Right-click on a databrowser plot
* Save Snapshot to File ...

About one time in ten, CSS will just hang and need killing. GUI redraws stop happening.

The code is in org.csstudio.swt.rtplot.internal.SnapshotAction.java and I think the offending line is in Plot.java at about line 440:

```
    /** @return {@link Image} of current plot. Caller must dispose */
    public Image getImage()
    {
        Image image = plot_image.orElse(null);
        if (image != null)
            synchronized (image)
            {
                return new Image(display, image, SWT.IMAGE_COPY);   // the problem is here...
            }
        return new Image(display, 10, 10);
    }
```

JConsole does NOT report a deadlock, but it does report that thread 'main' is blocked by 'Image@... owned by: Worker-0'. Worker-0 is waiting on org.eclipse.swt.internal.Lock@....

A possible solution is to put the image data fetch on the UI thread, but this may negate the benefit of having the bulk of the process inside a background job.

Here is a related bug that suggests that this is Linux-specific: https://bugs.eclipse.org/bugs/show_bug.cgi?id=265265

Finally, I realise @kasemir is working on a JFX version that will not be affected.  We do need some fix or workaround (whether that be as simple as disabling this feature) for using databrowser 2 on Linux.

kasemir:
If we remove the `synchronized`, the background thread might just then create a new image and call `dispose()` on the one we're holding.

What might work is this non-blocking approach that's aware of the image getting disposed:
````
    public Image getImage()
    {
        while (true)
        {
            Image image = plot_image.orElse(null);
            if (image == null)
                break;
            try
            {
                return new Image(display, image, SWT.IMAGE_COPY);
            }
            catch (Throwable ex)
            {
                // If image was disposed, try again
                // Otherwise give up
                if (! image.isDisposed())
                    break;
            }
        }
        return new Image(display, 10, 10);
    }
````

willrogers:
I'm not sure I fully understand.  If the image is disposed, are you confident that another not-disposed image will come along to replace it?

kasemir:
Yes. The image is replaced like this:
1) New one created
2) 'image' variable updated to the new image
3) If there was an old image, it's disposed

We synchronized to assert that the image which we received doesn't get disposed while we're looking at it.
[...]

PR #1998, opened 2016-12-13
Databrowser: avoid deadlocks taking snapshots.

willrogers:
Creating SWT Images on background threads can cause deadlocks:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=265265
Here just try again if the cause of exception was image being
disposed, since it will be refreshed.

Discussed in #1988.


berryma4:
Is this done.
(I understand this is not easy to test)

willrogers:
@kasemir to review.

kasemir:
I think this is it, thanks!

Issue #1999, opened 2016-12-13
Select default archive engine in search view if configured


willrogers:
In the data browser preferences you can choose default archive engines for each archive URL.

It would be helpful if the archive search view selected these by default, since (in our case at least) you just have to manually select the top one in the vast majority of cases.

kasemir:
If none is selected, it searches all of them, so you don't have to select any in your case.

But that may be more obvious if by default it started out with all selected.

willrogers:
Done.

PR #2000, opened 2016-12-13
Databrowser: select default engine for URL if selected.

willrogers:
If a default archive engine is specified in the preferences, select it
by default in the archive search view.

See #1999.

### DIFF
--- a/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/search/ArchiveListGUI.java
+++ b/applications/databrowser/databrowser-plugins/org.csstudio.trends.databrowser2/src/org/csstudio/trends/databrowser2/search/ArchiveListGUI.java
@@ -27,6 +27,7 @@ import org.eclipse.jface.viewers.ArrayContentProvider;
 import org.eclipse.jface.viewers.CellLabelProvider;
 import org.eclipse.jface.viewers.ISelectionChangedListener;
 import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.viewers.TableViewer;
 import org.eclipse.jface.viewers.TableViewerColumn;
 import org.eclipse.jface.viewers.ViewerCell;
@@ -63,6 +64,8 @@ public abstract class ArchiveListGUI
 
     /** Archive servers */
     private final ArchiveServerURL[] server_urls;
+    /** Default archives per server if selected */
+    private final ArchiveDataSource[] default_archives;
 
     /** Initialize
      *  @param parent Parent widget
@@ -70,6 +73,7 @@ public abstract class ArchiveListGUI
     public ArchiveListGUI(final Composite parent)
     {
         server_urls = Preferences.getArchiveServerURLs();
+        default_archives = Preferences.getArchives();
 
         createGUI(parent);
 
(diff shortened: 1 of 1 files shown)


berryma4:
(I like this!! Thank you!)
@kasemir okay?

kasemir:
When nothing is selected, it searches all of them.

That works for us, since we have two and by default search both.
Should also work for you with just one, since it searches that already.

But I agree that the UI is misleading.

-> select _all_ instead of the first one by default?

willrogers:
Maybe we have a bit of a strange setup, but we have something like 50 engines.  Unless I'm mistaken, sending a query to all of these would be prohibitively slow.

Then there's a master index, which indexes all of the others.  This is the default.  But we need to select it each time.  So selecting all by default doesn't really work for our setup.

kasemir:
We used to have ~80 when using the older Channel Archiver. That's one reason we stopped.

What now?
I have ~2 archive data sources and really want to search all of them by default.
You have many more and would like to default to the first one.

I could update my config so that the "most likely" data source is on top.
If that's selected by default, will my users notice that it's only searching that one, which is often sufficient, but if nothing is found, will they think about selecting the _other_ one?

Does this need to be a preference:
By default, select all, first, none?

willrogers:
Christ, it's difficult writing software for more than one person, isn't it?

I agree that my correct behaviour is not right for Kay's use case.

What do you use the default data source for?  Is that for if you just 'Add PV' or open a databrowser from a context menu?  If so, does this mean that you sometimes miss?

kasemir:
Yes, it's for each "Add PV". In addition, there's a config option to always use the default data sources for each channel, ignoring those in the *.plt file.
We used that as we transitioned the archive setup. We updated the default sources, checked the option to always use the default sources, so existing *.plt files that had old data sources then automatically used the new ones.

These settings are not used for the search, because there the idea was that you can specifically decide what you want to search.

willrogers:
Our situation right now is a usability problem.  My proposed change would be a usability problem for Kay.

You can have more than one default archive data source.  How does that work?

willrogers:
It looks to me like a PVItem can have more than one ArchiveDataSource.

Could Add PV search all the default archives?  This would help useability.

Then Kay could have both archives as default and the GUI would select them both.

kasemir:
Yes, I think this pull request is good.

Maybe that's obvious to everybody, but note there's a difference between the data server URLs and the archive data sources that they provide.
The preference "urls" list the data server URLs, for example

jdbc:mysql://localhost/archive
pbraw://somehost/...
xnds://localhost/archive/cgi/ArchiveDataServer.cgi

Each data server then provides a list of archives. For each RDB URL, that's just one. For the channel archiver, it's at least one (was ~80) in our case.

Each PV has a list of archive data sources that consist of a URL and a numeric key (and a name that's displayed, but not used to get data), for example

RDB|1|jdbc:mysql://localhost/archive
Vacuum (last week)|10|xnds://localhost/archive/cgi/ArchiveDataServer.cgi 
Vacuum (last year)|11|xnds://localhost/archive/cgi/ArchiveDataServer.cgi 
RF|12|xnds://localhost/archive/cgi/ArchiveDataServer.cgi 

The "archives" preference sets default archive data sources for new PVs, and also for all PVs when "use_default_archives=true".

The search view queries all "urls" for their data sources and lists them all.
This pull request will select all the default data sources in the list.
That'll work for DLS, where the default is just one data source, and they want to search just that by default, yet still allow selecting others.
It will also work for SNS, where the default is two archive data sources, and we'd like to search both, yet would still allow de-selecting one if that's desired.

berryma4:
merge then?

kasemir:
When porting to the udpated databrowser, I decided to select _all_ default data sources instead of just the first one. Makes no difference to DLS with just one, but makes more sense if you have >1 default data sources.

https://github.com/kasemir/org.csstudio.display.builder/commit/4593c09f7ec70ae38fa5b018a081fd95df10d0d6

Issue #2002, opened 2016-12-13
Linking Container OPI File property shall support macro


rosarp:
In order for the navigation pane to load dynamically only the mimic part of the
canvas, it is required that the linking container shall support macro.

This way, the navigation will update the macro ${OPI_FILE} and refresh the
canvas.
If the macro is inherited from the parent, then it does not work. 

For example:

1. Create A.opi. Add action button ACT_BUTTON.
2. Add Behavior->Actions->Open OPI
File Path = B.opi
Macros = {MIMIC_FILE = B_Mimic}
3. Create B.opi & B_Mimic.opi
4. The macro is used in B.opi in the linking container to load the mimic.
File path = ${MIMIC_FILE}.opi
5. Run A.opi & click on ACT_BUTTON
5. It will load B.opi but will fail to load B_Mimic.opi as the macro inherited from the parent A.opi is not used propeprly by the linking container to load the OPI.

To fix this issue, I propose following changes.

Mainly method fillLinkingContainerSub @@ -495,13 +511,19 @@
needs to be modified such that it fills runtimeMacros to resolve dynamically the opi path. 

Rest of the code is mostly plumbing to pass on macrosInput of parent opi from OPIRuntimeDelegate to XMLUtil.

berryma4:
It's best to share a branch, or make a pull request.
This code section is a touchy subject.

berryma4:
@shroffk, @willrogers, @kasemir 
Comment?

kasemir:
I agree that an embedded display with file path set to "$(FILE)" should resolve macros.
Note that it should NOT react to changes in the macro: Cannot update the macro at runtime and then have the embedded display widget re-load its content.
But should be OK as shown in the example when opening a new display:
Button 1 passes FILE=file1.opi to some display, which then ends up using $(FILE) for the path of an embedded widget -> It shows file1.opi.
Button 2 passes FILE=file1.opi to the same display --> Now it shows file2.opi.

For what it's worth, this does work in the display builder where most properties intrinsically support macros.

The pull request looks OK, but then we thought that before when it comes to embedded displays and macros.
Do you understand why the existing 
````
 String resolvedPath = MacroUtil.replaceMacros(path.toString(), s -> macroMap.get(s));
 path = ResourceUtil.getPathFromString(resolvedPath);
````
isn't sufficient, why a second replacement using
````
[...]

PR #2005, opened 2016-12-13
Linking Container OPI File property shall support macro.

rosarp:
Macros should be supported in OPI navigation.

Fix for #2002


applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/runmode/OPIRuntimeDelegate.java
```diff
@@ -136,18 +136,21 @@ public void init(final IWorkbenchPartSite site, final IEditorInput input) throws
                     inputStream = run_input.getInputStream();
                 displayOpenManager = run_input.getDisplayOpenManager();
             }
-            else
-                inputStream = ResourceUtil.getInputStreamFromEditorInput(input);
-
+            else {
+            	inputStream = ResourceUtil.getInputStreamFromEditorInput(input);
+            }
```
berryma4: Some white space issues

berryma4:
@shroffk I need a reviewer.  (and since you depend on this feature heavily ....)

shroffk:
I  do not have a well defined set of test outside the ones that exist in the opi-regression-test repository.
I will try to expand those screens...

The most important use case for us is that the Macros defined on the linking container widget itself should have the highest priority.

PR #2015, opened 2016-12-21
Databrowser: handle misconfiguration and log error.

willrogers:
See #1999.

If there is a problem with configuration of the default data sources, the Archive Search View won't render properly.  However, in other cases the same exception is unhandled as well.  So better to log a warning at the time and continue than to throw an unhandled exception.


willrogers:
Some whitespace errors have crept in from a previous commit, and this pull request is failing on those errors.

Issue #1937, opened 2016-10-24
Print action from opi runtime results in cs-studio crash


shroffk:
This issue can be reproduced on linux machines only. 

```
The program 'Cs-studio' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadDrawable (invalid Pixmap or Window parameter)'.
  (Details: serial 42818 error_code 9 request_code 53 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

nickbattam:
@shroffk Which OS/window manager(s) are you seeing this problem with?

I can print from an opi open in the OPI Runtime (right-click on opi) on my RHEL6 / GTK+2 machine using both the 4.3 and master branches with no error.

berryma4:
I can reproduce this on the 4.3.x branch, linux, gnome desktop.

shroffk:
Debian 8 with Gnome and KDE
with export SWT_GTK3=0 
So hopefully this should mean that we are still using gtk2

kasemir:
My setup is similar to Nick: RHEL 6.8 , and Help/About/../Configuration shows
`org.eclipse.swt.internal.gtk.version=2.24.23`.

Any more detail on the crash? Is there a stack trace that includes print-related methods which then end in some GTK/X call, and that receives the BadDrawable error?

shroffk:
There isn't a nice sack trace to look at, the abrupt exit from the jvm sort of makes it difficult to capture some of this information....

org.eclipse.swt.internal.gtk.cairoGraphics=false
org.eclipse.swt.internal.gtk.useCairo=false
org.eclipse.swt.internal.gtk.version=2.24.10
 
Later this week I will be able to provide more details..

berryma4:
I'm gtk version 2.24.25.
I can't seem to get a backtrace on the error.
The gdk_x_error break point never gets hit.
using:
```
GTK_DEBUG="all" GDK_SYNCHRONIZE="1" gdb /usr/lib/cs-studio/cs-studio

(gdb) break exit
(gdb) break gdk_x_error
(gdb) catch throw
(gdb) r -debug -pluginCustomization=debug.ini
```
I get a break at exit, but nothing for the gdk_x_error.
[...]

PR #2019, opened 2016-12-22
o.c.opibuilder: lengthened catch in PrintDisplayAction #1937

berryma4:
#1937


berryma4:
This should fix the issue

Issue #2017, opened 2016-12-21
Databrowser autoscales to all buffered data, not visible data


willrogers:
If you set autoscale on a databrowser Y axis, it will autoscale on all data contained in the `PVItem`.  However, this often contains historical data that has scrolled off the left-hand side of the plot, meaning that the autoscale no longer appears sensible.

Could it be changed so that it autoscales on the data visible in the plot?  From poking around I think this would be possible in `PlotProcessor.java`, by passing the x-axis range to `determineValueRange()`.

I can do this if you like.

willrogers:
The change would look something like the pull request I just submitted.  It is a behaviour change etc so needs approval.

kasemir:
Looks good to me.
Yes, it's a change, but one that matches the expectation of users.
Autoscaling based on some samples that aren't visible wasn't a good idea.
Thanks!

willrogers:
Done!

PR #2026, opened 2017-01-03
Databrowser: autoscale only to data visible in the plot.

willrogers:
The previous behaviour was to autoscale according to all the data in a
trace, even if that had scrolled off the left-hand side of the plot.

@kasemir see #2017.

### DIFF
--- a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
+++ b/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
@@ -49,11 +49,13 @@ public class PlotProcessor<XTYPE extends Comparable<XTYPE>>
         this.plot = plot;
     }
 
-    /** Submit background job to determine value range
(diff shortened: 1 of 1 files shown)


kasemir:
I've also implemented that for the updated databrowser.

Note that the value axes can have an inverted range, like 100 .. 0 where the "lower" end is larger than the "upper" end. In that case, 
````
if (item.getPosition().compareTo(x_range.getLow()) > 0 &&
    item.getPosition().compareTo(x_range.getHigh()) < 0)
````
fails.

This is unlikely for the X axis, specifically in the time-based data browser plot, but possible for the generic use of the RTPlot.

On the other hand, 
````
if (range.contains(item.getPosition()))
````
works in any case, so that's what I used in the JavaFX version.

**Comment by kasemir** [comment]
There's a problem with this for 'setpoint' type data that rarely changes.

Assume a PV was at 1 a week ago. Today it was changed to 2. I'm looking at a plot for today.

Would like to see the range 1 .. 2, but autoscale only considers values within the range of the plot, so it autoscales to 2 +- some tiny epsilon.

Fix: Autoscale has to use a range that includes the last sample just before the start time, based on 
```
start_index = PlotDataSearch.findSampleLessOrEqual(data, start_time);
```

PR #2025, opened 2016-12-28
4.5 patches

shroffk:
As requesting, creating a pull requests for patches from the 4.3 branch
#2016 #2001
Please delete branch after merging.

PR #2028, opened 2017-01-04
Whitespace

willrogers:
This gets rid of the whitespace problems that have been causing builds to fail.


berryma4:
https://github.com/ControlSystemStudio/cs-studio/pull/2025 also has the fix, ... but we had a problem yesterday that builds were doubled in time (TravisCI timeouts after 50mins)

willrogers:
Looks like this one passed - you could merge this and restart some of the other Travis builds to see if the problem is resolved.

Issue #2016, opened 2016-12-21
CS-STUDIO -> Utilities -> Search Logbook - the LogViewer Perspective is no more opened


rosarp:
@shroffk 

When opening search logbook, Log Viewer Perspective is not opened 

In Menu select :
CS-Studio -> Utilities -> Search Logbook

https://github.com/ControlSystemStudio/cs-studio/commit/e341dd334da154ac335dee865256b761c84e76b8#diff-4ca2cd8fe847f2d695265396691bef9c

Perhaps decoupling actions is still pending.
But, to have perspective opened on menu action, below changes are required.

```
diff --git a/applications/logbook/logbook-plugins/org.csstudio.logbook.viewer/src/org/csstudio/logbook/ui/OpenLogViewer.java b/applications/logbook/logbook-plugins/org.csstudio.logbook.viewer/src/org/csstudio/logbook/ui/OpenLogViewer.java
index a093f26..81d6185 100644
--- a/applications/logbook/logbook-plugins/org.csstudio.logbook.viewer/src/org/csstudio/logbook/ui/OpenLogViewer.java
+++ b/applications/logbook/logbook-plugins/org.csstudio.logbook.viewer/src/org/csstudio/logbook/ui/OpenLogViewer.java
@@ -44,7 +44,7 @@
         LogViewer.createInstance();
     }
     try {
-        //workbench.showPerspective(LogViewerPerspective.ID, window);
+        workbench.showPerspective(LogViewerPerspective.ID, window);
     } catch (Exception ex) {
         ExceptionDetailsErrorDialog.openError(
             HandlerUtil.getActiveShell(event),
```

Can we have this change so that, perspective is opened when search logbook menu is clicked?

shroffk:
Yes, you are right - this was something that slipped through the cracks.
I have re enabled the search logbook functionality to open the log perspective.

Also cleaned up the shift plugins.

Issue #2001, opened 2016-12-13
Alarm time since event context menu is empty


rosarp:
The context menu from an alarm shows first a clock icon for the time since
event: the elapsed time since the alarm happened.

But the time is empty.

Then clicking on the clock icon, the popup will show the time when the alarm
first occurs and the incomplete text "hh:mm:ss".

Perhaps Duration.between(start, end) is used wrongly in below commit.

https://github.com/ControlSystemStudio/cs-studio/commit/4804cb0ccfab06fe1eaff4d48c9ce8cce32c559e#diff-8d58f04efc6394964b546c3b169ae4f3R86

Same might have happened here..

https://github.com/ControlSystemStudio/cs-studio/commit/4804cb0ccfab06fe1eaff4d48c9ce8cce32c559e#diff-376a49bfb020c842093d3dd84482646dR136

rosarp:
@shroffk 

Can you please check this?

shroffk:
Thanks, 
I fixed the argument order and added a unit test at least for the NotifierUtil

shroffk:
addressed #2021 and #2025

PR #2030, opened 2017-01-05
#2025 Removing all references to guava

shroffk:
With the new java 8 stream api no need to use google


shroffk:
@berryma4 Could I include the rest of the pull patches to this PR or should they each be contributed individually?

berryma4:
#2016 #2001
You can fix more than more issue in a pull request, when they are clean commits (like you did).
But ... just small ones, the pull request needs to be easily tested.

PR #2031, opened 2017-01-05
In case of build failure on travis,search and print all log files #2025

shroffk:
In case of failure this configuration as suggested by @berryma4 should provide more useful information in the travis CI log

Is this acceptable
 - [x] @kasemir 
 - [x] @willrogers 
 - [x] @berryma4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,4 @@ script:
 
 after_failure:
  - find ./ -type d -name "surefire-reports" -print0 | xargs -0 -I {} find {} -iname "*.txt" -type f | xargs cat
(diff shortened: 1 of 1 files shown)


kasemir:
Only concern is that this will dump all logs, and then you don't know which log it is.
Would it make sense to look for 'configuration/*.log' instead of just '*.log'?
Or use something like this that echoes the name of the log file that's printed?
```
find . -type f -name "*.log" -exec "echo {}; cat {}" \; 

```

shroffk:
Added command to print the file name before the log file
find . -type f -name "*.log" -print0 -exec cat {} \;

willrogers:
Yes, always a good idea to try and clarify the output I think.

PR #2033, opened 2017-01-09
Opibuilder: optionally use pixels to specify font height.

willrogers:
For discussion - I haven't tried to make this perfect since it may be necessary to alter the strategy.

In particular, some widgets have additional configurable fonts that would need the additional property.

See #1990.


berryma4:
thank you for this!

berryma4:
Closing for #2035

PR #2035, opened 2017-01-11
1990 font pixel 2nd version

willrogers:
@kasemir @berryma4 another attempt.

As Kay presumably realised, this got fairly complicated.  However, I think this now has the behaviour that we want.  The remaining questions are:

* Have we caught all of the old examples where the resizing behaviour happened?
* Is this properly backward compatible and are the defaults correct?

See #1990.


berryma4:
This looks great!  Thank you!
@shroffk do you want to look at this first, before merging?

willrogers:
I think we might find a few things to correct in this, but it might be we have to merge and find them through testing.

Would it be worth running the opi regression tests?

berryma4:
If you like, we can push the opi regression tests to the release (which is after this PR, and the diirt config PR).  Before master is marked for frozen, the regression tests need to pass.

willrogers:
Yes, that makes sense.

Issue #2024, opened 2016-12-23
Removing warnings


claudio-rosati:
Eclipse Neon is showing 5732 warnings (for all the projects in CS-Studio).
Some of them are described in #2023.
It would be good revising the code/configuration and removing all these warnings.

shroffk:
+1
Although I think this has to be dealt with 2 different approaches, while some warning and errors can be addressed  by fixing code there are other that might need to be removed by configuring eclipse preferences.

willrogers:
@claudio-rosati have you read this page?  The part about Tycho configurators may reduce some of the warnings you're seeing.

Anyway, please show details of some of the warnings and we can see if some of them can be cleared up.

claudio-rosati:
Yew @willrogers, I've read the page on wiki, and everything was already done.

The attached [CR-5732-warnings.txt](https://github.com/ControlSystemStudio/cs-studio/files/701376/CR-5732-warnings.txt) file contains all the warnings I have. I hope you can figure out how to reduce them.

willrogers:
I'm generally keen to clean up warnings.  These mostly look genuine, so we should decide how we go about cleaning up.  

The current way we deal with these things is using the checkstyle rules: see https://github.com/ControlSystemStudio/cs-studio/blob/master/build/cs-studio-jenkins-checkstyle

We currently have a few which class as errors (resulting in a failed build) and a few which are just warnings.  An example of what we could do is:

* Remove all the unused imports reported
* Upgrade the warning to error for unused imports

Then the build would fail if anyone tried to commit an unused import, so we'd never get any more.  This would remove 439 of your warnings.  Some of the warnings are more difficult.

After a few false starts, this worked very well for the whitespace problems.

Thoughts @berryma4 @kasemir @shroffk ?

willrogers:
Although a lot of the errors in your file come from `org.csstudio.mps.sns`, which is not in the main CS-Studio repo so I don't think is particularly relevant.

claudio-rosati:
Thank you @willrogers.

When I was at SNS I added the org.csstudio.sns to Eclipse. I've then removed from the Project view but everything was still in the Navigator one.

Removing it and playing with the Problems view filters (allowing to fix all the warnings in the ESS product and in the diirt project I'm working now), brought me down to 3097.

willrogers:
Can you post another file with the remaining errors?

claudio-rosati:
Here the file:

[...]

PR #2037, opened 2017-01-12
2024 cleanup

willrogers:
See #2024.

This creates a new error in checkstyle, if there is an unused import.  It should have cleared up any remaining unused imports, so hopefully it will build.  I don't think it's controversial not to allow unused imports in future.


berryma4:
thank you!

PR #2043, opened 2017-01-17
Local enum pv init 2042

### LINKED ISSUE #2042: vtype.pv: Erroneous warning about initial value for VEnum
When using a local Enum PV
````
loc://demo<VEnum>(0, "One", "Two", "Three")
````
with initializer for initial value and labels several times in a display, this will result in warnings that the PV was once created with `(0, "One", "Two", "Three")` and once with `("One", "Two", "Three")` because the initial enum index is removed from the initializer list.

kasemir:
Fixes #2042


kasemir:
I see the test failure, will fix.

kasemir:
The o.c.vtype.pv.test now passes. Overall travis then stops after ~50 minutes because the build takes too long.

berryma4:
TravisCI offered to extend the time out to 70 minutes.  I'm also looking at using the target file again, because it caches the dependency resolution (instead of repeating at each module).

PR #2020, opened 2016-12-22
Data sources pref pages.

claudio-rosati:
The new Data Sources preference pages are completed and tested.
I would like @shroffk and @kasemir (at least) to give them a try.
I've already identified what to do to remove also the EPICS preference page and update the code using it. I would like to do it tomorrow.

PR #2049, opened 2017-01-23
As discussed using orbit provided versions of apache.commons.*

shroffk:
PR #2020 
@claudio-rosati


claudio-rosati:
Thank you @shroffk.

claudio-rosati:
@berryma4, can I merge?

PR #2065, opened 2017-02-03
Vtype pv prefs 2020

kasemir:
Updates vtype.pv to use the new diirt preferences #2020

Part of PV layer roadmap #1429


willrogers:
Could this be ported to 4.4?  Our 4.4 product uses the new Diirt preferences, but still has the EPICS preference page pulled in by the core base feature.

berryma4:
I was also thinking about this.
The two different preferences is a bit confusing in 4.4 (now that you can see two preference uis).

berryma4:
@kasmir, could this be cherry picked to 4.4?  Or, was there more work done?

kasemir:
As long has you have the org.csstudio.diirt.util.core.preferences plugin, on which the vtype.pv would depend with this update, you can cherry pick it onto other branches.

PR #2067, opened 2017-02-03
Pv tree 2064

### LINKED ISSUE #2064: PVTree handling of missing PVs and 'expand errors'
The PV tree shows when a PV is in alarm, but it doesn't specifically indicate missing PVs.

In this screenshot, the highlighted INPE has a missing PV.
It is shown just like the also highlighted INPH, which has a constant value:
tree1

Expected: Missing PVs should show invalid/disconnected alarm.

Especially for large trees, the 'expand only errors' button in the toolbar should expand the tree so that nodes in alarm are shown.
It currently seems to stop at the top node in alarm:

tree2

Expected (or at least more practical): Expand paths to all nodes in alarm.

kasemir:
Fixes #2064


kasemir:
Closing because by now I have a larger update of the PV tree to fix the code structure. Will create new pull request for that

PR #2071, opened 2017-02-06
Pv tree fixes, JFX

### LINKED ISSUE #2064: PVTree handling of missing PVs and 'expand errors'

kasemir:
Fixes #2064 :
 * Highlights disconnected PVs,
  * Identifies constant links
  * 'Expand tree branches in alarm' now works
  * Separates model from (SWT) view
  * Add JavaFX version of view
  * Add standalone application based on the JavaFX view
 * Add option to read fields as "field$", i.e. long string, to avoid the errors shown in screenshot where a long link text runs in the DBR_STRING length limitation


berryma4:
This is great!!  Thank you!

kasemir:
It is indeed interesting.
For now there's  both the SWT and JavaFX implementation because only the SWT one supports the context menu, while the JavaFX one can run standalone.
The JFX version seems more responsive when expanding or collapsing the tree.
Need to measure the performance for basic updates, if the values in the tree change.
In any case, it was good to detangle the model (connect to PVs, follow links, ..) from the representation (SWT, JFX, ..).

PR #2075, opened 2017-02-08
Adding Fault application to logbook module

shroffk:
An application to exploring the use of third party javafx libraries.
@berryma4 - please comment on the appropriate location


berryma4:
The build seems to be failing though.

shroffk:
@berryma4 I don't understand how it fails to find the branch from which this PR originates?
I ensured that the branch actually exists.

berryma4:
travisci probably doesn't like that your branch name doesn't follow the collaboration rules (issue_.*)  :)

berryma4:
It is strange though, permissions?

willrogers:
It looks like it's trying to fetch the branch from the ControlSystemStudio repo when it's on shroffk.

I don't know why, that's a perfectly normal thing to do.

shroffk:
I will recreate this PR

berryma4:
How strange this is, would you mind pushing the branch to CS-Studio, then pull request from there.

shroffk:
I thought a new PR would clarify the confusion but it still seems to be looking for the branch in the main repo instead of my fork.

As far as I am aware I did not do anything different with this branch.

PR #2084, opened 2017-02-13
Adding Fault application to logbook module

willrogers:
@berryma4 @shroffk this should be exactly the same as Kunal's.


willrogers:
... and same problem.

PR #2096, opened 2017-02-15
Connection widget issue 2095

### LINKED ISSUE #2095: Connection between widgets has routing issues
Mainly connection between widgets have following two issues.

1. Connection between widgets in different linking containers issue when the route is changed
Scenario:
> A.opi has two objects and one connector connecting them. The route of the connector is manually changed and fixed.
> B.opi has linking container which points to A.opi
Observation:
When B.opi is opened, the bendpoints of connectors from A.opi are at absolute position.
Thus, if linking container starts at (x=0, y=200), the bendpoints still stays at original location.
But, it is required to be relative to new position of linking container.

2. Connection between widgets is first displayed incorrectly and quickly refreshed properly
Same scenario as above, with B.opi having linking container and it points to A.opi.
It happens that routing is updated many times which causes routes to be displayed incorrectly first and then quickly refreshed.

rosarp:
Fixes #2095

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/LinkingContainerEditpart.java
@@ -253,6 +253,7 @@ public class LinkingContainerEditpart extends AbstractLinkingContainerEditpart{
             conn.setLoadedFromLinkedOpi(true);
             if(conn.getPoints()!=null)
                 originalPoints.put(conn, conn.getPoints().getCopy());
+            conn.setScrollPane(((LinkingContainerFigure)getFigure()).getScrollPane());
         }
         if (originalPoints != null && !originalPoints.isEmpty()) {
             //update connections after the figure is repainted.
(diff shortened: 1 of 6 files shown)


berryma4:
@rosarp could you add an opi to the https://github.com/ControlSystemStudio/opi-regression-tests repo; that shows the current failings, and then that this pull request fixes it.
We have difficulties testing linking containers, and we should add tests for behaviors expected.

berryma4:
you might need to rebase (I squashed your last commit).  Because this is a linking container change, it would be nice if you could also squash this to one commit.

shroffk:
@berryma4 @rosarp 
Sorry for the delay - is there an associated regression test in the repo?

Also, should I run the regression opi screens or have they been tested already.

mvitorovic:
Hello. As we're taking this over and this fix not being merged is blocking my other merge #2188, and I have a feeling that this issue is still not working correctly, can you please tell me what to do to have both pulled?

According to current experience, the fixed points routing connection is not working correctly in embedded OPIs, which I need to fix anyhow.

mvitorovic:
Ok. Looking at all this.... This code was a bit buggy and fixing those bugs actually resulted in redoing/undoing some of the changes in this pull request.
Maybe the best thing moving forward would be to abandon this pull request. The final pull request ( #2187 and #2188 ) will contain the full working FixedPointConnection and all related files:
- the router, 
- the fixes to the LinkinContainer and 
- everything that needs to be addressed for the fixed point connection to display poperly in all instances.

I'll also take care of the tests as well.

Does this sound reasonable?

berryma4:
thank you!

Issue #1860, opened 2016-08-22
BOY - Set line jump for connectors


utzeln:
User requirement (CRYO): there is one feature that would increase the “user experience” for HMI developers: automatically implement the gap when two connectors cross. Doing this manually is quite a burden.

I would propose to add new properties to the connector to set the line jump properties:
- line jump style: arc   | gap   | square   | 2 slides  
- add line jump to: none (will remove the line jumps) | horizontal lines   |  vertical lines  

- size of the line jump: 10

utzeln:
![line_jump](https://cloud.githubusercontent.com/assets/4686181/17848047/b3dab204-6851-11e6-87e4-33cc9249c649.png)

berryma4:
I like this a lot.  But, we should make sure it would also be supported in the new display builder @kasemir .

kasemir:
For the display builder, there's currently no plan from our side to invest development time in connectors. I don't consider it practical to turn the control system display editor into a schematic design tool where crossing wires can connect (with dot marker) or not (with arc | gap | square | ..) with auto-routing. 
Still, if somebody else wants to implement connector support for the display builder, fine.
But note that its model cleanly separates embedded (Linking Container) widgets from their content. The model is completely unaware of the embedded content. The content of an embedded display is a "black box". The embedded window runtime loads the content as a separate model. So there won't be any connectors from the inside of one embedded display to the inside of another embedded display.

utzeln:
"So there won't be any connectors from the inside of one embedded display to the inside of another embedded display" - it is unfortunate as it is one of ITER requirement.

kasemir:
I don't see a robust way to support it. The content of an embedded display is unknown. It can even change at runtime. So how can a wire connect to a specific widget inside an embedded display?

utzeln:
For now, the source and target for a connector is a succession of widget WUID which gives the path <wuid1>_<wuid2>_<wuid3> is using “_” as separator. So if a connector links 2 valves symbols embedded in grouping containers and linking containers, the source and target paths provides the full chain.

One consequence is that if the user delete a grouping container, then the connector cannot be displayed and generates an exception.

Then, our users are developing "super HMI objects" library where a valve symbol is represented by multiple PVs/statuses in a grouping container that can be linked in other displays. Once the library is published, widget WUID cannot be changed nor their hierarchy, otherwise the connectors between these widgets won't work anymore. In this case, we are managing manually the WUID and make sure they do not change.

[...]

PR #2097, opened 2017-02-15
Boy set line jump for connectors 1860

rosarp:
Adds line jump for connectors. #1860


kasemir:
No change. For the display builder, there's currently no plan from our side to invest development time in connectors, let alone ones that approach a schematic design tool with auto-routing.

Compared to BOY, the display builder already has better support for creating and editing the points of plain polylines. If you can place those on top of embedded content in the parent displays, you'll likely achieve the same optical end result in most cases. 

If somebody else wants to implement connector support for the display builder, a requirement would be that the inside of embedded displays (formerly known as Linking Containers) remains a "black box". There won't be any connectors from the inside of one embedded display to the inside of another embedded display, because the embedded content can change, orphaning connector end points. With this in mind I'm worried that `get/setParentDisplayModel` was added to the BOY display model, interlinking the embedded model even further.

Still, I'm OK with this pull request for those who plan to continue using BOY.

berryma4:
I'm not a big fan of adding too many features to BOY, that won't be the direction of the Display Builder. (making conversion and user expectations difficult)

kasemir:
Well, the way linking containers are merged into the overall BOY display model instead of trading them as a black box was, in my opinion, a bad design decision for BOY, but it's there.
ITER stated before that they depend on connectors linking into specific widgets inside linked containers. So we can't remove that from BOY, and ITER is of course allowed to continue to use and maintain BOY. Doesn't mean that a new tool must adopt the same design decision, and nobody is forced to update.

berryma4:
I'm going to squash the merge, so a confined change.

PR #2100, opened 2017-02-16
Fixed errors in manifest blocking Eclipse from building the plugin

claudio-rosati:
Yesterday I lost almost all .classpath and nothing more compiled.
I've fixed all of them but the org.csstudio.opibuilder.converter plugin suffered of a bad MANIFEST.mf having 2 non-existent packages exported.


kasemir:
Looks like those o.c.opibuilder.converter.prototype packages haven't existed for over a year.
@willrogers , @mfurseman, you know more about the converter, can you check?

willrogers:
I actually don't understand this bit.  My best guess is that the `prototypes` directory contained or contains some code used to autogenerate some of the Java classes used by the converter.

If you notice, there is a `prototypes` directory containing both some `.java` files and (odd) some `.class` files.

I expect you could also fix this by an entry in your `.classpath` file as follows.  This is what I have:

```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
	<classpathentry kind="src" path="prototypes/"/>
	<classpathentry kind="src" path="src/main/"/>
	<classpathentry kind="output" path="target/classes"/>
</classpath>
```

The entry with `prototypes/` in may fix the problems.

As to what the 'right' answer is here, I'm not really sure.

willrogers:
In fact, if I delete the `.classpath` file then import the `org.csstudio.opibuilder.converter` Maven project in Eclipse, Eclipse works out what is going on and creates the `.classpath` file correctly.

kasemir:
I think I get it.
The file `build.properties` contains
```
source.. = prototypes/,\
           src/main/
```

When doing the File, Import, Maven, existing maven project, it must parse that and create a .classpath file with the two source entries:
```
	<classpathentry kind="src" path="prototypes/"/>
	<classpathentry kind="src" path="src/main/"/>
```

Claudio somehow managed to delete many of his auto-generated .classpath files, and when he manually added a .classpath file, he didn't have that "src" entry for "prototypes"

-> We don't need this pull request, the MANIFEST.MF is correct as is.

claudio-rosati:
Thank you all for the comments.
There are still few things I don't understand: why Eclipse is so fragile and, more important, how to do the import @kasemir did.

I mean, I have to delete the damaged plugin from Eclipse (Project Explorer: right-click then Delete, without removing the contents from disk)? The import it again into the right working set?

PR #2108, opened 2017-02-16
travis: remove cat of tests

berryma4:


berryma4:
This "fixes" the issue, .. we now have errors.
I'm waiting on @shroffk whitespace fixes to rerun, and make sure it passes now.

willrogers:
I think this doesn't work because there is another error in the script, so it's failing too early.

See #2110.

PR #2112, opened 2017-02-17
Add .classpath file for opibuilder.rcp.

willrogers:
This allows explicitly adding access rules to allow access to JavaFX
and so removing Eclipse warnings.

Since .settings files are in .gitignore this is an exception and needs -f to add to the repo.  So this is just a suggestion, but I think it may be a good idea.

See #2024.

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/.classpath
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.rcp/.classpath
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+    <classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
+        <accessrules>
+            <accessrule kind="accessible" pattern="javafx/**"/>
+        </accessrules>
+    </classpathentry>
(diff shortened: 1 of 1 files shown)


willrogers:
There is one `.classpath` file committed at present, but I think that is just an oversight:
```
applications/opibuilder/opibuilder-plugins/org.csstudio.simplepv.test/.classpath
```

willrogers:
See also #2032.

kasemir:
For now doing the same with the display builder:
In general, .classpath files are ignored by git because they're auto-created upon Import/Maven/.. into the IDE.
Some plugins, however,  have a custom .classpath file in git to add the `<accessrule ..javafx`, and those .classpath files are specifically excepted in https://github.com/kasemir/org.csstudio.display.builder/blob/master/.gitignore .

It's not perfect: On Import/Maven/...,  the Tycho Configurator will replace those .classpath files. The Team (git) support will indicate changes to the otherwise pristine sources, and you need to Team/Replace/HEAD to get the .classpath files with `<accessrule ..javafx`.

For now I prefer doing this (i.e. replace-with-HEAD after each import into the IDE), to minimize warnings.
The access rule is not really necessary, though. Without the rule you get warnings in the IDE, and the auto-completion of imports won't list the JavaFX classes, but otherwise the code still compiles and executes in the IDE.

PR #2138, opened 2017-03-02
Disable keyboard traversal for most remaining widgets.

willrogers:
@nickbattam this seems to work for everything except for the XYGraph, which I can't figure out right now.


willrogers:
Sorry, wrong destination repo.

Issue #2086, opened 2017-02-13
How to set diirt preferences from plugin_customization.ini


kasemir:
The `vtype.pv` has recently been updated (#2065) to use the new diirt preferences (#2020).
This again allows configuring all EPICS settings from a `plugin_customization.ini` file,
either built into the product plugin or provided via the `-pluginCustomization` command line.

There's also again a preference GUI for the Channel Access settings, but it does not show the correct values.

**Example 1**
```
# plugin_customization.ini
org.csstudio.diirt.util.core.preferences/diirt.home=/tmp/bogus
org.csstudio.diirt.util.core.preferences/diirt.ca.addr.list=my_ca_gateway
```

_Result:_
`DIIRTPreferences.get().getString(ChannelAccess.PREF_ADDR_LIST)` returns "my_ca_gateway", vtype.pv can connect.
Preference GUI complains about missing diirt home (understandably) and shows "localhost" (wrong).

![dp1](https://cloud.githubusercontent.com/assets/1932421/22887982/25b95af6-f1d2-11e6-9d29-f00f960effd7.png)

_Expected:_
Preference GUI should show "my_ca_gateway".

**Example 2**
Trying the "_default_" variant of the preferences since that was mentioned in #2020:
```
# plugin_customization.ini
org.csstudio.diirt.util.core.preferences/diirt.default.initialized=true
org.csstudio.diirt.util.core.preferences/diirt.home=/tmp/bogus
org.csstudio.diirt.util.core.preferences/_default_.diirt.ca.addr.list=my_ca_gateway
```

_Result:_
`DIIRTPreferences.get().getString(ChannelAccess.PREF_ADDR_LIST)` returns "localhost", so the plugin_customization.ini value is not used at all.

**Example 3**
```
# plugin_customization.ini
# /tmp/ diirt exists,
# /tmp/diirt/datasources/ca/ca.xml contains addr_list="test"
org.csstudio.diirt.util.core.preferences/diirt.home=/tmp/diirt
org.csstudio.diirt.util.core.preferences/diirt.ca.addr.list=my_ca_gateway
```

_Result:_
`DIIRTPreferences.get().getString(ChannelAccess.PREF_ADDR_LIST)` returns "my_ca_gateway", vtype.pv can connect.
Preference GUI no longer complains about missing diirt home (good) yet still shows "localhost" (wrong).

[...]

PR #2143, opened 2017-03-07
Fixes for issue 2086: plugin_customization.ini not honored when displaying Preferences dialog.

claudio-rosati:


core/diirt/diirt-plugins/org.csstudio.diirt.util.core.preferences/src/org/csstudio/diirt/util/core/preferences/DIIRTPreferences.java
```diff
@@ -909,6 +970,110 @@ private static void forceDelete ( final File file ) throws IOException {
...
+                default:
+                    sb.append(c);
+            }
+
+        }
+
+        return sb.toString();
+
+    }
+
```
nickbattam: Java8 includes a function to do this encoding: https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html

claudio-rosati: Not working!
I need ' ' be converted to %20 not to '+'. This is why I had to do it myself.

claudio-rosati: Moreover, the "file:/C:/a/b/c/d..." will have also all ':' and '/' charachter encoded, another thing that I don't want.
This is not and encoding of a fragment, this is the encoding of the whole URL string in a way that the Eclipse mapper will understand.

claudio-rosati:
Please wait before merging... I found a thing that is noe broken

claudio-rosati:
@willrogers 
Good to know. I'll do it.
I don't like warnings, so being that I know what I've done, automatically I've suppressed it.

claudio-rosati:
OK, now everything is OK, and if @kasemir agrees on the fixes for #2086 the merge could be done.
Thank you for the patience waiting this fix.

core/diirt/diirt-plugins/org.csstudio.diirt.util.core.preferences/src/org/csstudio/diirt/util/core/preferences/DIIRTPreferences.java
nickbattam: I see. It would be useful to add that note in the function comment. 
An enthusiastic refactoring in a few years might make the same false assumptions I had and swap it out.

claudio-rosati:
Yes you're right.
Too many interruptions made me forget it.

core/diirt/diirt-plugins/org.csstudio.diirt.util.core.preferences/src/org/csstudio/diirt/util/core/preferences/DIIRTPreferences.java
claudio-rosati: Yes you're right.
Too many interruptions made me forget it.

willrogers:
I just built this version and the pre-defined preferences are correctly updated in the GUI.

One last question, and it's probably not important.  I don't have a preference defined for diirt configuration directory.  It defaults to @user.home/configuration/diirt.  Then there is a warning in the GUI because it doesn't exist.  Should it default to nothing, since it is not required any more?

Thanks for all your work on this.

willrogers:
It doesn't like the preference empty, either ...

claudio-rosati:
@willrogers 
The problem is validation: I can remove the warning message, meaning that if then someone will chose a wrong directory no message will be displayed either.
Generally I don't like how this messages are displayed in the dialog, so I would be more than happy removing it. I can write a warning in console instead.
If there is a general agreement on that it will be an easy a quick change.

willrogers:
Could the validation allow an empty string?

claudio-rosati:
Validation checks if the folder exists and if it has the right DIIRT structure (in terms of files and folders inside). So an empty string is no valid either.

What I can do is:

- showing a warning dialog if you chose a new "wrong" directory using the **Browse…** button;
- putting in RED the "DIIRT Configuration Directory" caption if the directory is invalid;
- no warning message in the Preferences dialog;
- a CONFIG log with the message previously displayed in the dialog;
- update the help accordingly.

Do you think it will be a good solution?

claudio-rosati:
@willrogers @kasemir @berryma4 @shroffk 
Can I proceed with my last suggestion?

@kasemir 
Have you tested if this pull request solve this issue?  (Will and I already tested it.)

willrogers:
This suggestion is OK for me.

claudio-rosati:
@willrogers can you please test the last changes?

Now an invalid (or empty) DIIRT home will display the **DIIRT Configuration Directory** caption in red. Moving the mouse over the caption will display the previous warning message in a tool-tip.

The warning message is logged too.

willrogers:
Works for me.

claudio-rosati:
@kasemir @berryma4 @shroffk
Can we merge it?

core/diirt/diirt-features/org.csstudio.core.diirt.feature/feature.xml
```diff
@@ -14,6 +14,12 @@
        Each library uses its own license.
    </license>
 
+   <requires>
+      <import plugin="org.eclipse.osgi"/>
+      <import plugin="org.eclipse.osgi.util"/>
+      <import plugin="org.eclipse.core.runtime"/>
+   </requires>
+
```
berryma4: are these needed?

kasemir:
OK with me.
The org.eclipse.* import may not be necessary, but at the same time my products have those included anyway, so I don't get any bloat from listing them in the diirt feature.

core/diirt/diirt-features/org.csstudio.core.diirt.feature/feature.xml
claudio-rosati: Yes I need them and being pre-loaded before anything els.

claudio-rosati:
Thank you.

Issue #2154, opened 2017-03-11
Memory Leak in Save & Restore


jbobnar:
Some listeners are not removed from a common service when a new save set is stored.

jbobnar:
Merged.

PR #2158, opened 2017-03-13
Dispose of selector when no longer needed by the RepositoryTree. (4.4.x)

berryma4:
#2154
Cherry pick from master


berryma4:
thank you!

PR #2163, opened 2017-03-15
DIIRT preferences overrides not saved if OK is pressed

claudio-rosati:
#2086
- Added missing statement when OK is pressed, so now Apply is no more
required being pressed before OK.
- Fixed the problem of not being able to revert to default if a DIIRT
xml file had a different value from the default one.


berryma4:
Thank you!!!
(side note: please don't include merge commits in the pull request)
I will cherry pick this to 4.4.x after the merge (I will do a squash commit to master to remove merge commits).

claudio-rosati:
@berryma4 
I really have not found a way to avoid the previous commits.
I don't understand if it happens because I'm working on a forked repository or not.
I just have my master branch (in the forked repository) in synch with the community one, I create a new branch, work on it, commit, push on the forked repository then I do a pull request.
So, I don't know what I'm doing wrong.
I've triend on Display Builder project to do squash using SourceTree, and at the end I've damaged my repository.
If you or some else can have an hangout with my to discuss the correct workflow to follow, I'm more than happy.
Claudio

willrogers:
Are you using the command-line for Git or are you doing everything through SourceTree?  I don't know that anyone else is using SourceTree so it's a bit difficult to figure out what is going on.

Git is undoubtedly a complicated tool, but now I've got the hang of it I wouldn't want to use another type of source control.

willrogers:
[This blog post](http://blogs.atlassian.com/2014/06/interactive-rebase-sourcetree/) talks about rebasing and squashing in SourceTree.  Try reading that page.

claudio-rosati:
Thank you Will, I'll do.

willrogers:
Confirmed bug 1.  I can't easily test bug 2.

I'm just building this branch.

claudio-rosati:
@willrogers 
I hope "confirmed bug 1" means that the bug was fixed.
Bug 2 is very tricky. I discovered it because my DIIRT configuration files were wrong, and the default setings were wrong too (having modified them to do some tests). So I don't think it should never happen in normal situations.

claudio-rosati:
@willrogers @berryma4 
After reading the article you pointed to me, I found the problem:

> Rebasing upstream commits
> Never rebase commits that have already been pushed. This can lead to dangerous and confusing situations.

This is the problem: I do some work on a branch, then I commit it and push to my forked repository. This will trigger jenkins on my repository, because I want to be sure everything is working fine before doing a pull request. If at the same time on the main repository someone has merged other branches, I'll synch them with my master branch to be sure my changes have no conflict. This will result in the multi-commits when I to a pull request.

When I tried to rebase before the pull request, I did it on commits already pushd, damaging the reposit. So It seems that I'm following a wrong workflow: probably I should never push until the development has completed, thus being able to perform the squash before pushing. Is that reasonable for you?

Issue #2107, opened 2017-02-16
java8 stream api as a replacement for gauva/google.collections


shroffk:
With a lot of the gauva features now part of the new java8 stream api we can try to refactor out the dependency to these google packages where possible.

kasemir:
Yes!

PR #2180, opened 2017-04-11
#2107 removing guava references in pretune application

shroffk:
#2107
Removing guava references


berryma4:
you have some failures in pretune on this.

Issue #2181, opened 2017-04-12
scan monitor: sort by selected column


kasemir:
Clicking on column header should sort on that column

PR #2182, opened 2017-04-17
Scan monitor sort 2181

kasemir:
Adds sorting to the scan monitor:
Clicking on table column header now sorts by that column.
Click again to change up/down direction of the sort.

#2181


kasemir:
Travis build failed because of 'rap-gef' URL. Merged up from master to get the updated 'rap-gef' URL.

PR #2183, opened 2017-04-17
vtype.pv: use diirt prefs 4.4.x

berryma4:
As stated in pull request #2065,
it seems like a good idea to have the single preference ui on 4.4.x.
I believe these are the commits needed.  But, if @willrogers could help test, that would be great.
(relates to #1429, but on 4.4.x)


core/base/base-plugins/org.csstudio.vtype.pv/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Basic PV that provides VType data
 Bundle-SymbolicName: org.csstudio.vtype.pv;singleton:=true
-Bundle-Version: 4.0.4.qualifier
+Bundle-Version: 4.5.0.qualifier
```
nickbattam: Is 4.5.x the right version for this feature on the 4.4 branch?

core/base/base-plugins/org.csstudio.vtype.pv/pom.xml
```diff
@@ -8,6 +8,6 @@
     <version>0.0.1-SNAPSHOT</version>
   </parent>
   <artifactId>org.csstudio.vtype.pv</artifactId>
-  <version>4.0.4-SNAPSHOT</version>
+  <version>4.5.0-SNAPSHOT</version>
```
nickbattam: Is 4.5.x the right version for this feature on the 4.4 branch?

berryma4:
@nickbattam good point, the versions wouldn't stomp each other because they are in different repos.  But, if someone wanted to be able to p2 update from 4.4 to 4.5, there might be a problem.  @kasemir maybe we could change the version to 4.4, or ... not change it?

kasemir:
I wanted to increment the vtype.pv version number because both the behavior (how to configure it) and the dependencies (now needs org.csstudio.diirt.util.core.preferences, which drags in common.io, common.lang3, ..).
Maybe it was wrong to jump to 4.5.x, misguided idea that it would look good to align with the overall 4.5 version of cs-studio? Still, going back to a smaller number on the master branch would be wrong now. So I'm for either keeping 4.5, because it's a copy of that same code on master, or maybe use 4.4, since it's a copy of that code but as it was compiled for the 4.4 branch.

berryma4:
I changed it to 4.4

nickbattam:
Thanks @berryma4

nickbattam:
@berryma4 I've tested this in our 4.4 build and it looks to be working correctly.

berryma4:
Thank you!

PR #2219, opened 2017-06-08
Timestamp ISO format & log formatting changed to TimestampFormats.

css-iter:
Added full formatting based on a preference. FULL, MILLI & SECONDS pattern are displayed with T and Z.
MILLI_PATTERN is used for log formatting.

### DIFF
--- a/applications/diag/diag-plugins/org.csstudio.diag.probe/META-INF/MANIFEST.MF
+++ b/applications/diag/diag-plugins/org.csstudio.diag.probe/META-INF/MANIFEST.MF
@@ -24,3 +24,4 @@ Require-Bundle: org.eclipse.ui,
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: Kay Kasemir <kasemirk@ornl.gov> - SNS
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Import-Package: org.csstudio.java.time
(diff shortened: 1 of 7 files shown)


kasemir:
Comments from hangout:

All agree that it's good to allow configuring the `csstudio/java/time/TimestampFormats.java` `DATE_PATTERN', `TIME_..`, `SECONDS_..`, `MILLI_..` and `FULL_PATTERN` from Eclipse preferences.

Change the pull request to have preferences for these patterns that provide the complete "yyyy-.." pattern, not just `format_datetime_with_T_separator=true`.

Default settings will reflect the current values.

This will add a dependency to the Eclipse preferences to a plugin that used to pure Java. Acknowledged.

kasemir:
This now also changes the `JMSLogMessage.DATE_FORMAT` from "yyyy-MM-dd HH:mm:ss.SSS" to `TimestampFormats.MILLI_FORMAT`.
The `JMSLogMessage.DATE_FORMAT` is used by the JMS messages that the alarm server and clients use to communicate. It's may also be logged and then used by reporting tools to summarize alarm system performance. This is thus not a time stamp that's displayed to end users which can be formatted as desired, it's a format used by software to communicate. Changing it can break existing setups.

Maybe this needs to be done in steps:

1) Update TimestampFormats.java to allow preferences
2) Update X to use TimestampFormats instead of some hardcoded value 
3) Update Y to use TimestampFormats instead of some hardcoded value 
4) ...

kasemir:
That way we can merge the TimestampFormats.java based on preferences right away, and then see which other pieces can be updated.

boryto:
OK,  we will revert the changes to `JMSLogMessage.DATE_FORMAT` and follow the suggestions.

boryto:
TimestampFormats.java added only to probe and logging plugins.

kasemir:
Thanks, this looks good to me for a merge.

PR #2226, opened 2017-07-03
Zoom linking container connections bug fix

css-iter:
When zooming, some connection anchors were moved out of bounds. This
caused connections from linking containers to disappear.


berryma4:
where is the issue ticket for this pull request?

berryma4:
Yes, all pull request need to reference an issue ticket.
Thank you!

PR #2227, opened 2017-07-03
Scan patch 2225

### LINKED ISSUE #2225: Scan 'patch' of value always uses string
When patching the value of a running scan, for example `Set ‘Mot:transX’ = 165` into a desired value of 170, the scan command is instead changed to `Set Mot:transX’ = “170”`, using a string as a value.
Because of Channel Access, that value as a string still works fine for the IOC, but if the command used comparison with a desired value, it'll fail.

kasemir:
Fixes #2225


kasemir:
Can we merge this? I'd like to deploy update at SNS beamlines.

kasemir:
Thanks!

PR #2230, opened 2017-07-05
vtype.pv pva table

kasemir:
Decodes pva://.. PVs of normative type table into VTable. Also tries to decode arbitrary structs into table as much as possible.

#1429


kasemir:
Thanks!

Issue #2235, opened 2017-07-11
Prompt for opening OPIEditor perspective is unreliable


willrogers:
I have a fix for this that we're testing over here at the moment.

willrogers:
Fixed.

PR #2239, opened 2017-07-19
2235 opieditor perspectives 4 4

willrogers:
See #2235


willrogers:
@berryma4 it would help us if this could get merged today.  Thanks!

PR #2241, opened 2017-07-20
Duration dialog and message history common Time-stamp format

boryto:
Common time-stamp formatting added to Message history, and timestamp helper  (notifier, logbook,  duration dialog, ...)


boryto:
Issue reference #2232

Issue #2217, opened 2017-06-06
Full ISO time-stamp formatting


boryto:
ITER has a requirement to use full ISO formatting (with letter T and zone).
To satisfy the requirement we propose an addition to TimestampFormats class. 

Additional setting will be defined: org.csstudio.java/use_full_iso_timestamp_formatting. 
When this is true, the patterns change to:
FULL_FORMAT =  "yyyy-MM-dd'T'HH:mm:ss.nnnnnnnnnX" -- ex. 2017-06-06T11:38:08.057000000Z
MILLI_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSX" -- ex. 2017-06-06T11:38:08.057Z
SECONDS_FORMAT = "yyyy-MM-dd'T'HH:mm:ssX" --ex. 2017-06-06T11:38:08Z

The preference will be false by default, so no change is expected for existing products.

boryto:
The org.csstudio.java.time.TimestampFormats has no Eclipse dependencies. If we want to read the preferences from the .ini file, the dependency on org.eclipse.core.runtime.preferences has to be added to the class.

Will this be OK?

css-iter:
#2219

boryto:
Created a pull request #2219

boryto:
Added formatting to Logbook. PR: #2223

css-iter:
Added to Databrowser, JMS monitor and Alarm displays: #2232

berryma4:
close?

boryto:
Yes, If there will be any new additions, we can create new one or reopen this one.

PR #2264, opened 2017-08-22
Added getTimeString overload which enables custom formatting

boryto:
#2217

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java
@@ -7,8 +7,8 @@
  ******************************************************************************/
 package org.csstudio.opibuilder.scriptUtil;
 
-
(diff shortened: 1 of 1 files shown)


applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder/src/org/csstudio/opibuilder/scriptUtil/PVUtil.java
```diff
@@ -87,211 +91,243 @@ public void partDeactivated(EditPart arg0) {
...
+     *
+     * @param pv
+     *            the Pv
+     * @param formatPattern
+     *            format pattern in the form of {@link DateTimeFormatter}
      * @return the timestamp in string.
      */
-    public final static String getTimeString(IPV pv){
+
+    public final static String getTimeString(IPV pv, String formatPattern) {
```
berryma4: This is the only real change, right?

boryto: Yes, the formatting seems to change the whole file, but this is the only change.

PR #2277, opened 2017-08-30
Combo: Ignore mouse wheel

### LINKED ISSUE #2276: BOY Combo should ignore mouse wheel
Our users tend to use the mouse wheel to scroll displays up/down.

If the mouse happens to be on top of a combo box, the scroll wheel events are captured by the widget and written to the PV.

This is unexpected. Writing to a PV, i.e. causing something to happen usually requires pressing "Enter" or deliberately clicking on something.

--> Combo should ignore the mouse wheel.

kasemir:
Fixes #2276

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/ComboEditPart.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets/src/org/csstudio/opibuilder/widgets/editparts/ComboEditPart.java
@@ -22,14 +22,15 @@ import org.csstudio.opibuilder.widgets.model.ComboModel;
 import org.csstudio.simplepv.IPV;
 import org.csstudio.simplepv.IPVListener;
 import org.csstudio.simplepv.VTypeHelper;
+import org.diirt.vtype.VEnum;
+import org.diirt.vtype.VType;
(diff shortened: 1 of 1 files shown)


kasemir:
Well, I found no way to suppress/ignore/abort the event. All I could do is recognize and then revert it.

nickbattam:
Setting `e.doit = false` prevents all actions linked to the change of selection occurring (i.e. updating the attached PV), but doesn't stop the displayed selection updating. 
If you cache the selectedIndex you can quickly reset that. Unfortunately the `SelectionEvent` doesn't include old index, so that would need to handled explicitly.

kasemir:
Well, what's your point? Do you want to cancel this pull request and send another solution for #2276?

nickbattam:
No point, I just raised it as a possible solution to your comment that you'd found no way to suppress the event.

kasemir:
OK. Well, as you noted, `e.doit = false` on the selection event still means that you need to patch the already updated UI, where the combo already shows the item that you didn't want to select after all.

Ideal would be a way to use `doit = false` on the mouse wheel event, but the `MouseWheelListener` receives a `MouseEvent` that has no `doit`.

With JavaFX, there's the concepts of filters which are called before the event handlers. So with the Display Builder, it would be possible to suppress the mouse wheel events. Alas, it turns out the JFX combo box doesn't have the problem in the first place.

Issue #2284, opened 2017-09-14
Can't delete Databrowser trace


willrogers:
Steps to reproduce:

* Delete PV name in text box and press enter
* You get a pop-up saying 'Cannot change item name...'
* Press OK

You can no longer delete that trace using the context menu.

willrogers:
Generated exception:

```
2017-09-14 15:26:07.983 WARNING [Thread 1] org.csstudio.swt.rtplot.undo.UndoableActionManager (execute) - Action failed: Delete Item
java.lang.RuntimeException: Not running 
	at org.csstudio.trends.databrowser2.model.PVItem.stop(PVItem.java:327)
	at org.csstudio.trends.databrowser2.model.Model.removeItem(Model.java:474)
	at org.csstudio.trends.databrowser2.propsheet.DeleteItemsCommand.run(DeleteItemsCommand.java:51)
	at org.csstudio.swt.rtplot.undo.UndoableActionManager.execute(UndoableActionManager.java:57)
	at org.csstudio.trends.databrowser2.propsheet.DeleteItemsCommand.<init>(DeleteItemsCommand.java:43)
	at org.csstudio.trends.databrowser2.propsheet.DeleteItemsAction.run(DeleteItemsAction.java:85)
	at org.eclipse.jface.action.Action.runWithEvent(Action.java:473)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:565)
	at org.eclipse.jface.action.ActionContributionItem.lambda$4(ActionContributionItem.java:397)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5227)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1340)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4561)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4151)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:1121)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:336)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1022)
[...]

PR #2286, opened 2017-09-15
Prevent changing DB pv name to an empty string.

willrogers:
This could get the plot into an inconsistent state where you cannot
delete the item.

See #2284.


kasemir:
Thanks!

PR #2288, opened 2017-09-15
Boy Text Input: Use control instead of display range for "Limits from PV"

### LINKED ISSUE #1910: Text Input Widget Malfunctions with PV Limits
Attn @berryma4 

Using the FRIB Debian package 4.3.2-7.

If the connecting PV for a `Text Input` widget has its `DRVH/DRVL` set, but no matching values in HOPR/LOPR, then arbitrary put-values result in the pv being set to zero.

To reproduce, simply create an `analog out` record served with `softIoc -d ai-test.db`, caput to the record to give it an initial value (say 5), caput it's DRVH to 10 (for example), and then try to use a Text Input widget to set the pv to 3. This will result in the pv being set to zero.

kasemir:
Fixes #1910


berryma4:
Thank you!

Issue #2291, opened 2017-09-21
Repainting connections from property change


boryto:
Regarding issue  #2281, there was a bad looped update in the PolylineJumpConnection. This fixes the issue
Repainting is added on some property changes, so connection repaints directly after property changes.

PR #2293, opened 2017-09-21
Fix for connection paint loop.

boryto:
This fixes the problem with connection gap and the issue with loop on paint method call. 
In ref to #2291


willrogers:
This isn't anything I've really looked at before, sorry.

It looks fairly harmless, adding a `repaint()` into an 'ignorable ui task'.

PR #2307, opened 2017-10-13
Respect DRVH/L in Knob and Slider widgets

### LINKED ISSUE #2305: ScaledSlider and Knob widgets ignore drive limits from PV
When the ScaledSlider widget is set to get limits from PV it sets the max and min scale limits to  UpperDisplayLimit (HOPR) and LowerDisplayLimit (LOPR). 

This update is done is a base class shared with a number of other widgets:
Monitor:- Guage, Meter, ProgressBar, Tank, Thermometer
Control:- Knob, ScaledSlider

For the 'monitor' widgets the DisplayLimits are probably the correct field to use. 

For the 'control' widgets should the max/min be set to the drive limits (DRVH and DRVL) rather than display limits to match behaviour in the TextInput (#1910) and SpinnerWidgets (#2302)?

nickbattam:
Introduces a distinction between 'control' and 'monitor' widgets in abstract class hierarchy at the Model level.

Introducing this distinction in the EditPart implementation, through properties or additional levels of abstract class was less clean.
 
Resolves #2305


tashasummers:
I like that all the control widgets are now going to use DRVH/DRVL as the ranges. 
For the monitor widgets, could there be an option to use the DRVH/DRVL values if HOPR/LOPR are both left as default 0?

Issue #2312, opened 2017-10-24
Archiver Appliance reader doesn't support access for retired PVs


nickbattam:
For display of PVs that are no-longer being archived, but do exist in the archive, the AA reader doesn't support the required API

Issue #2309, opened 2017-10-20
Updated version of pbrawclient jar file


willrogers:
This is used for the Archiver Appliance archive reader plugin.

The version in the repo is 0.0.4; we needed a capability that was just added in 0.0.8.

willrogers:
Completed.

PR #2313, opened 2017-10-24
Update to support getData for retired PVs

nickbattam:
#2312

### DIFF
--- a/applications/archive/archive-plugins/org.csstudio.archive.reader.appliance/src/org/csstudio/archive/reader/appliance/ApplianceArchiveReader.java
+++ b/applications/archive/archive-plugins/org.csstudio.archive.reader.appliance/src/org/csstudio/archive/reader/appliance/ApplianceArchiveReader.java
@@ -244,6 +244,8 @@ public class ApplianceArchiveReader implements ArchiveReader, IteratorListener {
      * http://domain:port/retrieval/data/getData.raw where /data/getData is
      * fixed and .raw identifies the MIME-type of the returned data.
      *
+     * To retrieve data for retired PV(not archived anymore), use retiredPVTemplate as suggested by Murali Shankar
+     * e.g.CSS Data Browser URL:http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate=templatePVName
      * @return data retrieval URL
      */
     public String getDataRetrievalURL() {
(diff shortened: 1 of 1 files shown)


berryma4:
Could you explain this a little (the url change) in a comment in the code.

k-vijayan:
Yes comment added now.

berryma4:
Maybe I'm not understanding the setup.  Are you suggesting that in the property file, instead of putting the retrieval address as: http://server:port_number/retrieval/, you put: http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate.
I don't understand how this is suppose to work.

k-vijayan:
Requirement:
We need both the Data Source:
1. `http://server:port_umber/retrieval`  - To retrieve archiving PV data from AA
2. `http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate=TemplatePVName` - To retrieve not archiving PV data (old PV's/retired PV's) from AA.
3. Use the latest pbrawclient.jar https://github.com/slacmshankar/epicsarchiverap_pbrawclient/releases/tag/0.0.8 which was updated in  #2309 

e.g. 
1/(data is from 2006 to 2009) BL-AL-LEVEL-01:HEIGHT is Retired PV to plot on Data Browser.

2/RetiredPVTemplate = SR02C-PS-Q2AF-01:ILKS (still archving in AA)

In this situation we create a new datasource
`http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate=SR02C-PS-Q2AF-01:ILKS` in DataBrowser Settings/Preference (Archive data source URL).

The log shows this is converted into the full URL to get the data from AA into CSS:
`org.epics.archiverappliance.retrieval.client.RawDataRetrieval (getDataForPV) - URL to fetch data is http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate=SR02C-PS-Q2AF-01:ILKS&pv=optimized_800%28BL-AL-LEVEL-01%3AHEIGHT%29&from=2009-02-01T15%3A35%3A00Z&to=2009-02-28T15%3A35%3A00Z`

berryma4:
This is the part I don't understand:
```http://server:port_number/retrieval/data/getData.raw?retiredPVTemplate=SR02C-PS-Q2AF-01:ILKS```
The template pv is used for the etl storage information (where to look).  Does the datatype, as you put in the notes, also have to match (boolean, double, array, etc).
How often would this property need to be changed?

k-vijayan:
Yes I think it should have the same data store and other properties as well.
For more information check Murali's updated guide/document about this retiredPVTemplate. 
http://slacmshankar.github.io/epicsarchiver_docs/userguide.html#retrieval_url_formula

berryma4:
Okay, so he suggests having a paused pv (template pv) to hold the storage information (pvinfo).  It seems the datatype does not matter, just the storage.
This is all seems odd.  Why couldn't we just have a boolean in pvinfo that says it is archived or live?
So, in the end, I see why you are doing it this way.  Sorry for the delay.

k-vijayan:
Thanks

Issue #2268, opened 2017-08-29
Various 4K hidpi issues


boryto:
On 4k screens,  issues with scaling of some components arise, specially when using bigger fonts. 
The first to be solved is the log entry dialogue:
![image](https://user-images.githubusercontent.com/5550006/29805631-74c1b9c0-8c7a-11e7-8200-b33e895c52ae.png)
Hide/Show details button almost hidden behind all other elements. 

This issue addresses all hidpi problems.

boryto:
With Neon release, Eclipse has native support for scaling images on hidpi monitors.   https://www.eclipse.org/eclipse/news/4.6/platform.php#swt-autoscale 
Scaling is realized with SWT scaling  - the basic principle is re-sizing existing icon images and taking  images with @2x or @1.5x (200% or 150% scaling) at zoom levels to get crisp images. 

While this looks good on most of the Cs-studio components, it represents problem on BOY opi's. Specially existing ITER opi's, which are already designed for 4K. The core of the problem is that ITER has already developed OPI's which are designed for 4K. This opi's already contain 4k images, and opi's are designed with borders over 3000px (width) . If we use native scaling,  images and borders are also scaled. This causes the existing opi's to look bad (over-scaled) when scaling to 150% or 200%.

One way to fix this issue would be to only disable this kind of scaling for boy, but this support is build direct into Image class, and there is no parameter on the Image or ImageDescriptor class to disable this scaling. There are also some icons (alarm table widget), that are used in boy and in alarm table. So it doesn't look so good. 

Seems that only reasonable way to deal with the problem is to create bigger icons, and make the preference to enable them.  We could use the same suffix for images @2x. If preference is true, @2x icon images would be used. If not, normal images would take place. So we have full control over icon images. This would require a helper class (as ImageUtil)  to be used at many places in code.

Any other ideas?

kasemir:
Can you attach screenshots of some section in your displays with older CSS vs. based on neon, on normal and 4K display?

boryto:
Attached screenshots are displays with no scaling, and with 150% scaling. When no scaling is used the Opi's look the same as on Mars. So the diffrence is only between not scaled and scaled. 

The editor icons on scaled screen look OK, but the OPI is to big, as the images and borders are bigger.

![editor_noscaling](https://user-images.githubusercontent.com/5550006/30058023-aff0b3d0-9228-11e7-861a-0c22d7413426.png)
[...]

PR #2315, opened 2017-10-26
Fix for the mixed arguments/results table in Service button dialog

boryto:
Tables were switched, so arguments were under results.


boryto:
#2268

shroffk:
Was this a result of the previous merge?

boryto:
Previous merge (05585afc59) was fix for the dialog, as before the dialog was totally messed-up.  This is additional fix, for the tables of results and arguments were flipped after the merge.  Not sure, but yes this, could be result of previous merge.

shroffk:
ok cool...I am fine with the merge

The form layout has behaved badly on past occasions on various linux windows managers. I think we should try to avoid it.

Issue #2320, opened 2017-11-14
Throttling updates when using the Scaled Slider Widget


mfurseman:
At Diamond we have IOC's that log when a PV is written on them. This introduces a delay in receiving the a caput. If these PVs are updated rapidly, such as with the scaled slider widget, then the caputs queue and it takes a long time for the PV to settle to the final value.

Is there any throttling build into Diirt to control writing of PVs, has anyone else experienced this problem?

PR #2323, opened 2017-11-15
Throttle the rate of updates when using the scaled slider widget

mfurseman:
This adds a timer that limits the rate of control updates to once per 100ms to address https://github.com/ControlSystemStudio/cs-studio/issues/2320.


mfurseman:
As far as I can tell, the only fireManualValueChange listener is contributed by the ScaledSliderEditPart, this sets the PV which also notifies the PVValueListeners. The only widget that provides a PVValueListener is the ArrayEditPart.

So for your point above, as the only action is to write to the PV, there is no reason to have this on the UI thread anyway.

Issue #2329, opened 2017-11-20
Queue of screens to open can get overloaded


willrogers:
When you're launching a bunch of screens from external commands, after a while the requests may no longer be honoured, or they can wait for a long time.

The reason for this is that we're using `WorkbenchAdvisor.eventLoopIdle()`, which (if a bunch of stuff is going on) can just stop being called, presumably because the event loop isn't idle enough to decide to call this method.

[This article](https://www.eclipse.org/articles/Article-RCP-2/tutorial2.html) from many years ago suggests that this isn't the best way to do background work, and suggests the jobs tutorial.

I've made a branch which is more responsive in our case, but needs review to make sure there isn't something I'm missing.

PR #2330, opened 2017-11-20
Don't use WorkbenchAdvisor.eventLoopIdle() to do background work.

willrogers:
The openDisplay jobs are being handled on background threads using asyncExec,
and the eventLoopIdle() method is not reliably called.

Not for merge just yet, needs review @kasemir @berryma4 @shroffk .

See #2329


willrogers:
Well, it's not caching the requests any more, it's just sending them straight to an `asyncExec()`.

willrogers:
OK, we'll put this in our nightly build and make sure it's working as expected.

Issue #2353, opened 2017-12-19
MEDM converter results in <wuid> entries which complicate file diffs


kasemir:
Mark Rivers noticed that converting the same *.adl file again results in a different *.opi file because of the unfortunate `<wuid>`.

Since hardly anybody needs those, especially not for files generated from *.adl inputs, they should be omitted in the converted files.

kasemir:
In addition, ADL ""related display" widgets are converted into MenuButtons. Better would be an ActionButton because that allows opening as replacement, new tab etc.

PR #2354, opened 2017-12-19
ADL converter: Remove <wuid> from output

kasemir:
#2353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_script:
 
 script:
  - java -version
-   # Make sure stdout is in blocking mode. If we don't, then conda create will barf during downloads.
+   # Make sure stdout is in blocking mode.
    # See https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959 for details.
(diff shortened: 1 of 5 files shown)


kasemir:
One Travis run fails with Java Heap Space, the other shows "Build: Success" and then ends with "cat: write error: Resource temporarily unavailable".
Don't see actual error

berryma4:
I'm going to change the build a little to help with this.
I'm going to deploy from org.csstudio.product, and change the travis files in maven-osgi-bundles, cs-studio-thirdparty, cs-studio to not deploy.
https://github.com/ControlSystemStudio/cs-studio/issues/2355

berryma4:
There seems to be an issue on TravisCI:
https://github.com/travis-ci/travis-ci/issues/4704#issuecomment-348435959

It would probably be fine to just remove the cat at the end.
But, to keep the cat *.log, we would need to add:
```python -c 'import os,sys,fcntl; flags = fcntl.fcntl(sys.stdout, fcntl.F_GETFL); fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);'```
to the travis file.

Issue #2374, opened 2018-01-16
Message history OPI widget


boryto:
A new OPI message history widget, wrapper around the existing widget is created.

To satisfy all desired functionalities, additions to existing message history are made:
-Ability to create new instances of message history widget.
-Selecting different columns for widget instances.
-Time stamp custom formatting.
-Option to query filtering that parses sql like syntax and sets the corresponding filters.

Filtering syntax is simple and only takes AND separator between columns, and = for setting the values.
For ex: Start_time=-1 hour AND End_time=now AND name=%CTRL%

SQL syntax was changed a bit, so we can actually limit to max messages not the max. properties shown.

PR #2375, opened 2018-01-16
Issue 2374

boryto:
New message history OPI widget .

Rrf: #2374


kasemir:
This changes the SQL to add a nested "SELECT ( SELECT ..)".
Have you been able to test that with MySQL?

boryto:
We have tested nested query directly on mySql (command line client) . We had no problems with nested queries.  

Example of tested query:
`SELECT m.id, m.datum, m.TYPE, m.NAME, m.SEVERITY, c.msg_property_type_id p, c.value FROM (SELECT * FROM message msg WHERE msg.datum BETWEEN '2017-10-18' AND '2017-10-19' AND msg.id IN ( SELECT message_id FROM message_content WHERE value LIKE '%pv%') ORDER BY msg.id DESC LIMIT 20) m, message_content c WHERE m.id=c.message_id;
`

Issue #2379, opened 2018-01-22
Start BOY window in full screen mode


boryto:
For now in cs-studio product, the following preference is setup:
org.csstudio.opibuilder/start_window_in_compact_mode=true

But there is not its equivalent to start window in full mode. And swtich off
the compact mode does not set it in full mode neither.

It is required to be able to configure BOY to run the screen in full mode with preference
eq.: start_window_in_full_screen_mode=true

There is also status line disappearing in compact and full screen mode. 
This should also be configurable via preferences.ini

PR #2380, opened 2018-01-22
Issue 2379

boryto:
In ref with: #2379 

Also the code which always resized and moved the cs-studio window in case there was only one
window open in runtime mode was removed.


boryto:
No, the hack is still needed. 
But we had some strange behaviour even with this hack, as it was not working in some cases.

PR #2388, opened 2018-01-31
Medm cvt 2387

### LINKED ISSUE #2387: MEDM converter tweaks
From Mark Rivers:

..conversions are really good..

Many of the issues we did find were due to problems in the medm files themselves.  medm tolerates text widgets that are too small to actually hold the text.  It also will display an enum value as a string even if the datatype of the widget is "decimal" rather than string.  I have fixed these problems in the medm files..

.. a few remaining issues ..

- When a widget has color mode "alarm" in medm the colors are NO_ALARM=green, MINOR=yellow, MAJOR=red.  In CSS we get NO_ALARM=black, MINOR=orange, MAJOR=red.  This leads to displays that are very difficult to read, particularly when NO_ALARM text is on a dark background.  Could we make the alarm colors in the autoconverted files match medm?

- Text graphics widgets with Right Justify are not right justified in CSS.  It looks like Left and Center justify are OK.  The same is true for text update widgets.

- Slider widgets do not convert well at all.  Note that the CSS widget is tiny, with unreadable text.

- When a related display with multiple displays has a label (e.g. "Multiple displays" in attached medm screen) that label does not appear in CSS.

If these problems were fixed I think that the conversion tool would be really excellent. The autoconverted screens would be sufficient in nearly all cases, with no need for tweaking.

kasemir:
Fixes #2387

### DIFF
--- a/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.server.test/src/org/csstudio/alarm/beast/server/NagTimerUnitTest.java
+++ b/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.server.test/src/org/csstudio/alarm/beast/server/NagTimerUnitTest.java
@@ -41,7 +41,8 @@ public class NagTimerUnitTest implements NagTimerHandler
     @Test(timeout=8000)
     public void testNagTimer() throws Exception
     {
-        final NagTimer nag_timer = new NagTimer(1000, this);
+        final long period = 1000;
+        final NagTimer nag_timer = new NagTimer(period, this);
         nag_timer.start();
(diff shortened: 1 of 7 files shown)


berryma4:
Keep getting a test failure on:
```[INFO] org.csstudio.alarm.beast.server.test ............... FAILURE [01:42 min]```

kasemir:
The `NagTimerUnitTest` expects a notification every second, checking for the first one after 1.1 seconds.

So if the test runs 0.1 seconds slow, it'll get the error.

I'll update the test to use longer times.

PR #2390, opened 2018-02-02
When reading font.def, use font_default_pixels_or_points

### LINKED ISSUE #2389: Honor 'font_default_pixels_or_points' preference in font.def
Fonts can be defined based on pixels or points, #2145
Current versions of BOY save fonts with this information.
When loading older *.opi files, the preference setting `font_default_pixels_or_points` determines how the font size is interpreted.

This preference setting is not, however, used when reading the font.def file.

kasemir:
Newer format font.def can then use "px" or "pt" on each font, but
default comes from the same preference that's used for all fonts.

Fixes #2389


kasemir:
Good point(!). Just like "px", and explicit "pt" now overrides the default from preferences.

kasemir:
The `AlarmNotifierUnitTest` fails in one of the builds in what looks like a timing related issue.
Unrelated to this pull request.

Issue #2412, opened 2018-03-28
Alarm Server: Write severity to PV


kasemir:
Some alarm handlers allow writing the severity of a node in the alarm configuration hierarchy to a PV.
ALH has the $SEVRPV for this purpose; SNS has some generated CALC record hierarchies that summarize the alarm state. Either way, there are existing operator interfaces and sometimes even IOCs that depend on PVs which summarize the alarm state of sub-tree in the alarm configuration hierarchy.

==> Add similar functionality to the alarm server.
To minimize changes, this would be implemented as a special form of "automated action,
for example "sevrpv:SomePV" would update `SomePV` with the alarm severity of the node.
Note this would be the `SeverityLevel`, i.e. ok, minor_ack, .., minor, .., undefined.

PR #2419, opened 2018-04-05
Alarm Server sevrpv: support

kasemir:
Implements  #2412 : 

Automated Action on Area, system, subsystem with detail `sevrpv:NameOfPV` instructs alarm server to write the severity of that node to a PV, very similar to ALH `$SEVRPV`.

In case of worries that the alarm server quits and thus will no longer update the PVs, a new `org.csstudio.alarm.beast/heartbeat_pv` preference can be set to the name of a PV that'll be periodically set to 1, usable for a heartbeat check.

Details are in the online help.

### DIFF
--- a/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/demo/heartbeat.db
+++ b/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/demo/heartbeat.db
@@ -1,0 +1,18 @@
+# Example for "Heartbeat PV"
+#
+# Alarm Server sets it to 1 every ~10 seconds,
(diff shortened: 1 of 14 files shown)


berryma4:
Thank you for the heartbeat addition!!

kasemir:
Yes, that might be useful on its own.

Set `org.csstudio.alarm.beast/heartbeat_pv` to a record that looks like https://github.com/ControlSystemStudio/cs-studio/blob/master/applications/alarm/alarm-plugins/org.csstudio.alarm.beast.configtool/demo/heartbeat.db

berryma4:
You could even make the record plc backed to a phone-dialer!
(Super safe!)

Issue #2401, opened 2018-03-09
Update to Eclipse 4.7 Oxygen


willrogers:
We want to update CS-Studio to build and run using Java 9. A prerequisite for that is to build against Eclipse Oxygen.

berryma4:
let's update to the new diirt release and then branch first.

willrogers:
Fine, there's no rush.

kasemir:
Eric, you make it sound as if changing diirt is easier than updating to Java 9 ;-)

berryma4:
master is open for updating to Oxygen.

berryma4:
I merged ControlSystemStudio/cs-studio-thirdparty, did you have a branch for ControlSystemStudio/cs-studio?

berryma4:
I made a pull request with the new urls.
We were able to remove efx, because it is now in the Eclipse Release Train (the oxygen url).

berryma4:
I've also updated the common product urls, and added the java 9 vmarg
```java
--add-modules ALL-SYSTEM
```

berryma4:
Also, with java 9, we should be able to use GTK3:
http://openjdk.java.net/jeps/283

berryma4:
tried upgrading to e(fx)clipse 3.0.0
I had issues with the diirt jars/MANFEST resolving javafx.*
I tried adding extraClasspathElements for javafx to the tycho compiler, but this didn't work.
Out of frustration, rolling back to e(fx)clipse 2.4.0
I still removed:
all javafx imports, org.eclipse.fx.javafx from all requires (not needed anymore)

claudio-rosati:
About e(fx)clipse... and JDK 11/OpenJFX 11

https://tomsondev.bestsolution.at/2018/08/04/supporting-openjfx-11-from-jdk11-onwards-in-efxclipse/

PR #2420, opened 2018-04-06
update to Eclipse 4.7 Oxygen

berryma4:
#2401 
I have updated the orbit link, and removed efx (efx was added to the Eclipse Release Train).
I didn't update RAP.  I saw they have an update, but I'm not sure how to test.


berryma4:
I need to remove references to org.eclipse.fx.javafx, and javafx imports.

PR #2432, opened 2018-05-16
Welcome16.png not existing anymore

claudio-rosati:
Updated Welcome menu icon to png version instead of gif, no more supported in Eclipse 4.7 Oxygen.

I've tested before the change the ESS and SNS product, and both have an exception because wrlcome16.gif was not found. Digging into org.eclipse.ui.intro.universal I've seen that that icon was replaced by a `.png` and a `@2.png` versions.


claudio-rosati:
@berryma4 
I'm not able to do it.
Each time I've tried to do it I get conflicts and error. I'm sorry but I cannot manage it. Maybe next time can we have an hangout so you can help me in doing it.

willrogers:
It is mysterious where all these merge commits come from. Is it SourceTree?

claudio-rosati:
I'm working on a forked repository, and locally with SourceTree, yes.

Issue #2437, opened 2018-06-05
Phoebus Integration


shroffk:
Create a plugin which provides a service which allows for the launching of cs-studio/phoebus.

shroffk:
@kasemir 
I have two options when it comes to including launch actions for various phoebus applications from cs-studio.

Option 1.
Add the launch action in the associated application plugin itself...i.e in the o.c.diag.probe
This would add a dependency to the o.c.phoebus.integration plugin and would need additional preferences to allow users to enable/disable including phoebus actions.

Option 2.
Create a new stand alone plugin which registers the actions. 
This would mean that the app launch action would not be in the same plugin with the old cs-studio applications...but it would be easy for people to control weather to include these actions in their product.

Which option should I add...I am leaning towards Option2. since it would be the lease disruptive although not the nicest way.

kasemir:
Yes, 2, for the reasons you mention.
That way, you can decide if you want to include 'probe' or 'launch-probe-in-phoebus', 'PV Tree' or 'launch-PV-Tree-in-phoebus', 'XXX' or 'launch-XXX-in-phoebus' in your site-specific product.
I also see value in not having to change the existing XXX code at all for this purpose.

shroffk:
Integrate (menu actions)
 - [ ] probe
 - [ ] databrowser
 - [ ] displaybuilder
 - [ ] alarm applications
 - [ ] channel applications
 - [ ] logbook applications

Integrate (context menu actions)
 - [ ] probe
 - [ ] databrowser
 - [ ] displaybuilder
 - [ ] alarm applications
 - [ ] channel applications
 - [ ] logbook applications

PR #2439, opened 2018-06-05
Issue #2437

shroffk:
Work in progress. Do not merge


willrogers:
These changes don't directly affect me so merge when you're happy.

berryma4:
Sounds like a consensus. Merge this, and new pull requests for new features.

claudio-rosati:
From a user point of view, what will appear on the CS-Studio UI?

willrogers:
This pull request introduced a bunch of whitespace errors that broke my build. I guess we don't have the Travis build at the moment to catch these.

berryma4:
My fault, yes it looks like the Travis/GitHub pull request stopped being
part of the "check marked okay". I'll look into this.

shroffk:
Sorry, 
It seems like with my installation of oxygen I did not have eclipse-cs plugin available bu default, I usually just carry over the preferences from my old workspace but this time I had to install it separately.

Even after it is installed and it seems to not apply to xml files...a resource described in stylesheet as one to be checked for tabs and whitespaces.
In the example below I added tabs and whitespaces and tried "check code with checkstyle" but got no warnings while there are clearly 2 issues.
image

I would hope the IDE warns me of these violations before they are committed and break builds...

Issue #2452, opened 2018-08-02
Indicator of the current time in databrowser plot


willrogers:
If you're panning around in the databrowser, sometimes you get a gap in data on the right-hand side. This could be:

* we are waiting for the databrowser to fetch data
* the area on the right-hand side of the plot is actually in the future, so there is no more data to show

It would be nice to have some visual indication of _now_ so you can tell if the data is fetched. This could just be a vertical line, or something like the right-hand side of the plot being shaded.

@skyfrench is looking at this at DLS, but @kasemir might also have thoughts.

willrogers:
Restricting what you can see on the graph does somehow feel a bit restrictive. Maybe you want to centre something that means you need a space on the right? Maybe not. Anyway, we've made a proposal in the pull request.

PR #2457, opened 2018-08-03
2452 data browser current time indicator

willrogers:
See #2452 

There are different ways this could be implemented (comments welcome @kasemir), but I think the effect is pretty good.

now

cc @skyfrench


berryma4:
This is great! Maybe you could make the dashed line lighter, and have the time axis after now match this lighter gray.

willrogers:
I think that's a good suggestion, but it depends on changes to RTPlot so I'd welcome @kasemir's opinion.

kasemir:
This is just for non-scrolling mode, right?
When scrolling, the right edge of the plot is at 'now', so the divider line isn't show. I'd hate to have a dancing line at the right edge of the plot.

As for what to show, a divider line doesn't mix well with grid lines.

I find simply changing the time axis label color like this best to indicate that we're off the normal time range:

now

willrogers:
I think we do currently have a dancing line on the right, but you're right we should turn that off.

I see your point about grid lines, but the grey numbers are a bit too subtle for me - I don't think people would notice them. A solid pale grey line? Another option is to lightly shade the part of the plot that is 'in the future'.

I think that this could be a very intuitive feature that people would understand immediately.

kasemir:
I like shading the right section of the plot:

now_bg

I'll add that to DB3 and the phoebus version.

For DB2, you might also consider adding an office assistant like this:

clippy

berryma4:
I like the shading a lot! Clippy could be an Easter egg preference.

kasemir:
OK, DB3 and Phoebus version now use the shading. Assuming this is generally useful, they have no option, the future is simply always shady.

kasemir:
Here's the corresponding change for DB2:
```
diff --git a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java b/applications/appunorganized/appunorganized-plugins/org.
index e17fd29..8a2dfd9 100644
--- a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java
+++ b/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java
@@ -47,6 +47,7 @@ import org.eclipse.swt.events.MouseMoveListener;
 import org.eclipse.swt.events.MouseTrackListener;
 import org.eclipse.swt.events.PaintEvent;
 import org.eclipse.swt.events.PaintListener;
+import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.graphics.GC;
@@ -678,6 +679,21 @@ public class Plot<XTYPE extends Comparable<XTYPE>> extends Canvas implements Pai
         }
 
         gc.setClipping(plot_bounds);
+
+        if (x_axis instanceof TimeAxis)
+        {
+            final int future_x = ((TimeAxis)x_axis).getScreenCoord(Instant.now());
+            final Color orig = gc.getBackground();
+            // Use light gray for bright background, otherwise dark gray
+            final Color shade = (background.getHSB()[2] >= 0.5)
+                ? new Color(getDisplay(), 200, 200, 200, 255)
+                : new Color(getDisplay(), 50, 50, 50, 255);
+            gc.setBackground(shade);
+            gc.fillRectangle(future_x, 0, area_copy.width - future_x, area_copy.height);
+            gc.setBackground(orig);
+            shade.dispose();
+        }
+
         plot_area.paint(gc, media);
 
         for (YAxisImpl<XTYPE> y_axis : y_axes)
```

willrogers:
Err... is clippy my not-as-subtle option? I think I'm happy with the shading.

willrogers:
> the future is simply always shady

The future's looking pretty shady to me.

kasemir:
I'm unclear about the future. Maybe slightly worried.
Clippy is simply a friendly assistant missed by everybody who knew him (her? it?). No hidden meaning.
Alas, I lack time to implement a useful clippy. May no longer be necessary now that users can talk to Siri, Alexa and the like.

willrogers:
Closed in favour of #2465

PR #2459, opened 2018-08-06
2453 data browser cancel search

willrogers:
This cancels more frequently.

We had another problem which probably wants a timeout on the XmlRpcClient calls, but as we hope to drop the Channel Archiver pretty soon we figured it was a bit tricky to fix.

cc @skyfrench


kasemir:
I think the bigger problem is that it's hard to actually cancel requests for data once submitted to the RDB or other data source. You're stuck in "get-the-reply()" with no chance to check if a cancellation has been requested. The ArchiveFetchJob starts another thread for that so that it can then 'cancel()' the reader, but that might not have any effect.

kasemir:
Ported to DB3 and Phoebus version

Issue #2455, opened 2018-08-03
On a Value Axis 'Add' the automatic axis naming can create a duplicate of an existing name


skyfrench:
e.g.

When you right click and call 'Add Axis' from the Value Axes tab the new axis currently takes the name Value N where N is the number of existing value axes + 1.

Therefore if you currently have:
Value 1, Value 2, Value 3
Then delete the first one, leaving you with:
Value 2, Value 3
The next newly added axis gets the name Value 3 (2 + 1), giving:
Value 2, Value 3, Value 3

willrogers:
Fixed

PR #2460, opened 2018-08-06
2455 data browser value axis naming

willrogers:
See #2455


kasemir:
Has been ported to DB3 and Phoebus version

PR #2465, opened 2018-08-07
Databrowser: shade the area of the plot that's in the future.

willrogers:
I slightly changed Kay's suggestion - it's a slightly lighter colour and the grid lines will be painted over the top.

See #2452 and #2457 

future


kasemir:
As for exact colors, the last 'alpha' argument in the color calls `new Color(getDisplay(), 240, 240, 240, 0)` is a relict of me trying to lay a translucent rect on top of the grid lines, as done in the JFX version. That didn't work out, since anything but 0 and 255 are ignored for the SWT colors. So you did the right thing by placing the gird lines on top.

willrogers:
Good point - I tried 0 to see if it made it translucent, but it didn't seem to. So I've removed the alpha values as they're not having any effect.

Issue #2467, opened 2018-08-08
DataBrowser 'Stagger' traces button does not disable autoscale


skyfrench:
When autoscale is switched on and the stagger traces button is clicked, the traces are staggered, but then immediately restored to their previous view as the graph autoscales the axes back. A call to stagger traces should switch off autoscale.

PR #2468, opened 2018-08-08
cs-studio #2467 (CSS-495): DataBrowser 'Stagger' traces disables autoscale

skyfrench:
If the 'Stagger' traces button is clicked on the DataBrowser, autoscaling is now turned off for all value axes. Previously any introduced Stagger was undone by the plot auto-scaling back to the originals afterwards.

### DIFF
--- a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/RTPlot.java
+++ b/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/RTPlot.java
@@ -25,6 +25,7 @@ import org.csstudio.swt.rtplot.internal.ToggleLegendAction;
 import org.csstudio.swt.rtplot.internal.ToggleToolbarAction;
 import org.csstudio.swt.rtplot.internal.ToolbarHandler;
 import org.csstudio.swt.rtplot.internal.TraceImpl;
+import org.csstudio.swt.rtplot.internal.YAxisImpl;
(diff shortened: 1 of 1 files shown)


kasemir:
Checked DB3 and Phoebus version; both already disable autoscale on 'stagger'.

While at it, is there a better proper English name for the 'Stagger' option?

Short for "an arrangement of things so that they don't overlap"

Even shorter for  "Arrange axes limits such that traces for axes are stacked on top of each other without overlap. For N axes, vertical space is divided into N sections of equal height. Limits of first axis are set so that its traces occupy the top section etc."?

"Stack"?

willrogers:
Sky just started at Diamond. She's not in my team, but she is helping me out with CS-Studio while I'm short-handed and her other responsibilities haven't yet ramped up.

kasemir:
Excellent. Welcome!

willrogers:
Before I knew what 'stagger traces' meant, I didn't think it was explanatory. Now I do know, I can't think of a better word...

kasemir:
"Distribute vertically"?
When I type "arrange so as to be parallel" in google, the links for the most promising sites lead to crossword solvers, access to which is blocked from my work computer because they're blacklisted as gaming sites.

claudio-rosati:
I like "Distribute Vertically"

willrogers:
I just talked to a user of this feature (@gilesknap). What he often does is move to a different timebase which might mean losing the traces from the current Y axes. He uses 'stagger traces' because it gets all of them on the screen, but what he sometimes wants is more like 'centre traces on all Y axes'. The stagger button is the same, but it separates the different axes to take up different parts of the vertical space.

Another point to note is that traces are only separated if they are using different Y axes.

Issue #2458, opened 2018-08-03
Relative start times (> 1 month) in the Data Browser not handled correctly by start/end time dialog box


skyfrench:
Relative start times > 1 month set by the start/end time dialog box are converted into days for display in the 'Start Time:' text edit. When the start/end time dialog box is re-opened, the time is displayed incorrectly because the day spinner takes a maximum value of 31.

e.g. to reproduce:
- Open a Data Browser plot
- Open the dialog [...]
- Set a relative start time of 1 year
- Accept the change; 365 days is displayed
- Re-open the time dialog
- The relative start is shown as 31 days in the spinner boxes, but 365 days in the start time test field

PR #2471, opened 2018-08-09
carry over days over 31 when updating the RelativeTimeWidget GUI

willrogers:
See #2458 

I don't know if this has changed in newer versions of the databrowser, but these start and end times are the source of a lot of confusion and bugs. I wonder if there's a better way of presenting this.

cc @skyfrench


kasemir:
DB3 uses the same SWT start/end time dialog, so your fix is for both.

In Phoebus, we have a JFX start/end time dialog:

time

It combines the relative and absolute times in one view, i.e. no more switching between tabs.
The active section is highlighted, as in the example there's a relative start time (2 years, ..) and a relative end time (now). If you clicked on the start date or time, the upper left section would become highlighted because you now have an absolute start time (the selected date & time) with a relative end (now).

In any case, I think the rollover is handled in the new version. Incrementing the months beyond 12 increments the years. Selecting 2 years stays as 2 years, not 2 * 365 days or the like.

kasemir:
Here's an example where both an absolute start and end time was selected by clicking on the 'Date:'.
The calendar which pops up shows the effective start..end date range.

time2

Issue #2466, opened 2018-08-08
Display more than one waveform in the WaveformView


skyfrench:
The existing WaveformView isn't very useful for comparing different waveforms. It would be useful if it was possible to:
- plot more than one Waveform (PV) in the view
- draw the waveform as a line (instead of scatter points)

PR #2474, opened 2018-08-13
#2466 Waveform View support Multiple Traces

skyfrench:
Implemented support for multiple traces in WaveformView
- Possible to select more than one PV for display
- Traces are now drawn as a line with no points

Modification made to handle events not coming in at the same time for different plotted PVs:
- the bottom slider remains a linear slider which scrolls through the samples available in the first plotted item (i.e. the single waveform display behaves as it does currently, and when there are multiple traces the first plotted behaves in the same was as it would were it plotted alone)
- the samples plotted for the other item(s) are then the preceding waveform(s) from the timestamp of the waveform sample in the first item

cc @willrogers


kasemir:
Pull request #2473 was submitted just before this one, and since it has been merged, there are merge conflicts on this one. Please check into using the existing MultipleSelectionCombo class, and the merge conflict. Otherwise I think this is a great idea, thanks!

willrogers:
I resolved the merge conflict and did the simpler list removal. The `MultipleSelectionCombo` sounds useful.

kasemir:
Yes, `MultipleSelectionCombo<ModelItem>` directly holds the model items, removes the need for most of `getModelItems()`.

kasemir:
You're right. With check boxes, you just tick or un-tick the boxes. With the multiselectioncombo, you need to Shift-or-Ctrl-Click until you get carpal tunnel syndrome.
I'm in favor of updating the multiselectioncombo to use combo boxes.
The phoebus version already uses combo boxes, which is probably another indication of their superiority.

skyfrench:
I have committed a modification to the MultipleSelectionCombo to use check boxes. Hopefully it's an improvement on what was there previously?

kasemir:
Does the MultipleSelectionComboDemo still work?

skyfrench:
> Does the MultipleSelectionComboDemo still work?
Yes, I believe it does (comparing behaviour before and after).

willrogers:
That looks nicer. Two comments:

* I don't seem able to get rid of the dropdown by clicking elsewhere (your previous version worked)
* Maybe it should say "- Select items -" or something when nothing is selected

Issue #2483, opened 2018-08-14
Data Browser archive data request type preference


skyfrench:
It would be nice to allow a user to specify a Data Browser default archive data request method in Preferences.

kasemir:
Not sure.
The data browser defaults to 'optimized', and what that means is up to the archive data source.
For the RDB, it will fetch N (default is N=800) min/max/average samples, unless there are fewer than 800 raw samples in the requested start..end range, in which case you get the raw samples.
This means you can look at 1 year, 1 month, 1 week of data and get the best possible overview, i.e. outline of min/max value range as well as average, while reducing the network traffic as well as client memory usage. As you zoom in, you eventually get the raw data.

Why would you want to switch to 'raw'?

1) The 'optimized' implementation is broken for your archive data source.  -> Fix it.
2) The user always wants to see the raw data. Yes, of course. But then the user will soon complain that fetching the data takes longer, redrawing the plot takes longer, and CSS aborts with out-of-memory. Not a practical approach.
3) There are edge cases where the plot for a channel simply looks odd when requesting optimized data. This is typically with sparse data, for time ranges around the threshold of automatically switching between optimized and raw data. Manually switching to raw makes the plot look better. Sometimes increasing the number of `plot_bins` for optimized data helps to avoid the situation. In theory, it should be similar to the screen width in pixels: In a plot that's 1000 pixels wide, showing more than 1000 data points doesn't improve the display. Makes more sense to show 1000 min/max/average samples from the 'optimized' request. But that's the theory, which in practice differs from reality. Not sure what to do about it, but again simply always using 'raw' doesn't seem like a good solution.

berryma4:
What might be a nice feature, would be to have an option in the preferences to set the plot_bins from the pixels available. This way, we wouldn't have to have a different config for different monitor resolutions.

willrogers:
OK, I'm behind this suggestion. We have a real problem, let's explore what the correct answer is. It might be that you cover it in case three.

We have a magnet at 0A. We turn it on, and it rapidly moves to 10A. In the final bin are values 7, 8, 9, and 10; the average is 8.5 A. It stays at 10A for multiple bins, then we turn it off. The next bin with values is 10 and 9; average 9.5A.

What we see is trace at 8.5A for the whole period where the magnet has actually been at 10A. This will induce (and has induced) alarm in the control room. Switching to 'raw' data gives the correct view and usually results in the user vowing always to switch to raw data.

[...]

PR #2486, opened 2018-08-14
Data Browser: Allow plot_bins as scaling factor for display width

kasemir:
#2483

Allows negative values for plot_bins to scale the display width,  i.e. -3 means: 3 times Display pixel width.


berryma4:
@willrodgers

willrogers:
I think this makes sense. What happens if there are multiple points for a pixel? Does it plot multiple times?

kasemir:
Just like 'raw', it plots those points, with some minor optimization: If a new point's x/y pixel is the same as the previous, it's skipped.

PR #2508, opened 2018-08-29
CSSTUDIO-824: JCA no more used

claudio-rosati:
Channel Access "mode' property removed in favour of an always set pure-java mode.


claudio-rosati:
OK, I've found them and reverted back.

PR #2510, opened 2018-08-30
Phoebus integration

claudio-rosati:
These are the changes made by Kunal and Claudio at ESS to have it properly working.

PR #2511, opened 2018-08-31
Phoebus integration #2510

claudio-rosati:
These are the changes made by Kunal and Claudio at ESS to have it properly working.

Clean PR.


berryma4:
Thank you for the clean commit

Issue #2448, opened 2018-07-25
SVG images not displayed in BOY


claudio-rosati:
Hello all,
I need help on this issue that, I've verified, appeared only since we moved from Neon to Oxygen (It was reported to me only now because BOY is not used everyday).

When opening am OPI containing an Image widget loading an SVG file, it is not displayed, and the following is on the log file:

```
!ENTRY org.eclipse.ui 4 0 2018-07-25 12:11:51.189
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.LinkageError: loader constraint violation: when resolving method "org.apache.batik.bridge.BridgeContext.getCSSEngineForElement(Lorg/w3c/dom/Element;)Lorg/apache/batik/css/engine/CSSEngine;" the class loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader) of the current class, org/csstudio/utility/batik/SVGHandler, and the class loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader) for the method's defining class, org/apache/batik/bridge/BridgeContext, have different Class objects for the type org/apache/batik/css/engine/CSSEngine used in the signature
	at org.csstudio.utility.batik.SVGHandler.buildElementsToUpdateList(SVGHandler.java:786)
	at org.csstudio.utility.batik.SVGHandler.<init>(SVGHandler.java:188)
	at org.csstudio.swt.widgets.symbol.SVGSymbolImage.loadDocument(SVGSymbolImage.java:313)
	at org.csstudio.swt.widgets.symbol.SVGSymbolImage.access$2(SVGSymbolImage.java:300)
	at org.csstudio.swt.widgets.symbol.SVGSymbolImage$2.runWithInputStream(SVGSymbolImage.java:280)
	at org.csstudio.swt.widgets.util.AbstractInputStreamRunnable.run(AbstractInputStreamRunnable.java:32)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:37)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:182)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4034)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3701)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1150)
[...]

PR #2513, opened 2018-09-03
Pin versions of Batik to avoid class load errors.

willrogers:
Eclipse Oxygen appears to provide both 1.7 and 1.8, which is causing
errors when loading SVGs.

See #2448.


berryma4:
Thank you!

Issue #2519, opened 2018-09-17
Multistate symbol widget doesn't show string values correctly


willrogers:
This seems to be a regression introduced by cd54400. If the PV associated with a multistate symbol widget is a string, the label no longer shows it correctly, while it did in 4.5.0 and before.

It seems like the following diff will resolve it:

```
diff --git a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.symbol/src/org/csstudio/opibuilder/widgets/symbol/multistate/CommonMultiSymbolFigure.java b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.symbol/src/org/csstudio/opibuilder/widgets/symbol/multistate/CommonMultiSymbolFigure.java
index a95d09c938..c4a50961a3 100644
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.symbol/src/org/csstudio/opibuilder/widgets/symbol/multistate/CommonMultiSymbolFigure.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.symbol/src/org/csstudio/opibuilder/widgets/symbol/multistate/CommonMultiSymbolFigure.java
@@ -226,6 +226,7 @@ public abstract class CommonMultiSymbolFigure extends Figure implements SymbolIm
                 statesLong.add(null);
             }
             statesStr.add(state);
+            index = statesStr.indexOf(state);
             statesMap.put(index, state);
             IPath path = findImage(index);
             if (path == null)

```

PR #2520, opened 2018-09-17
2519 multistate

willrogers:
See #2519.


willrogers:
"The job exceeded the maximum log length, and has been terminated."

Issue #2524, opened 2018-09-20
Allow databrowser warning message suppression


willrogers:
We have a problem in requesting optimized data from the data browser, described in #2483.

Somewhat related, there are two warning messages that come up every time you use an option for each run of CS-Studio:

* hiding a trace
* requesting raw data

I realise that the problem with requesting raw data requires another fix, but there are reasons to hide traces (e.g. you want to clear up your plot view for some reason, but you don't want to delete a trace that you have configured).

It would be more friendly to have an option to suppress these messages via a preference:

![hide-trace](https://user-images.githubusercontent.com/2440055/45827204-da449980-bced-11e8-889e-784dc87bc142.png)

PR #2525, opened 2018-09-20
Allow the option to suppress data browser warning messages.

willrogers:
For each launch of CS-Studio, you get fresh messages about hiding traces
and requesting raw data. This commit allows suppressing these messages
via a preference.

See #2524


berryma4:
passed tests, but hit travis maximum log length.

PR #2526, opened 2018-09-26
2466 waveform view

willrogers:
This is a number of tweaks to fix and improve the behaviour introduced in #2466.


kasemir:
I'm not 100% sure about this:
```
final int idx = new TimeDataSearch().findSampleGreaterOrEqual(samples, sliderTime);
sample = samples.get(idx - 1);
```
The located sample index could be 0 (valid, first sample).
But `idx-1` would then result in an error.
For databrowser3, I was that happen when adding two waveforms and then quickly trying to operate the waveform view, so I used `samples.get(idx);` without the `-1`.

willrogers:
Let me think...

willrogers:
OK, I know what I was thinking but in practice yours works better.

willrogers:
I think this one is ready to merge.

PR #2529, opened 2018-10-01
Update travis.sh to Maven batch mode

berryma4:


berryma4:
I saw that when I googled around.

claudio-rosati:
A stronger approach will be `-Dorg.slf4j.simpleLogger.defaultLogLevel=warn`

berryma4:
Much cleaner!

Issue #2531, opened 2018-10-05
css crashes when converting an adl file


MarkRivers:
When I try to convert this adl file CSS crashes (I have added a .txt extension so Github would upload it).
[13IDD_LVP_T25.adl.txt](https://github.com/ControlSystemStudio/cs-studio/files/2450563/13IDD_LVP_T25.adl.txt)

This file contains the complete output of CSS when I try to convert that file:
[error.txt](https://github.com/ControlSystemStudio/cs-studio/files/2450590/error.txt)

This appears to be the fatal error:
```
Handling: text
org.jdom.IllegalNameException: The name "1" is not legal for JDOM/XML elements: XML names cannot begin with the character "1".
        at org.jdom.Element.setName(Element.java:207)
        at org.jdom.Element.<init>(Element.java:141)
        at org.jdom.Element.<init>(Element.java:153)
        at org.csstudio.opibuilder.properties.MacrosProperty.writeToXML(MacrosProperty.java:101)
        at org.csstudio.opibuilder.properties.ActionsProperty.writeToXML(ActionsProperty.java:146)
        at org.csstudio.opibuilder.persistence.XMLUtil.widgetToXMLElement(XMLUtil.java:96)
        at org.csstudio.opibuilder.persistence.XMLUtil.widgetToXMLElement(XMLUtil.java:104)
        at org.csstudio.opibuilder.persistence.XMLUtil.widgetToXMLString(XMLUtil.java:145)
        at org.csstudio.opibuilder.adl2boy.ADL2BOYApplication.convert(ADL2BOYApplication.java:111)
        at org.csstudio.opibuilder.adl2boy.ADL2BOYApplication.checkThenConvert(ADL2BOYApplication.java:103)
        at org.csstudio.opibuilder.adl2boy.ADL2BOYApplication.start(ADL2BOYApplication.java:66)
        at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:196)
        at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:134)
        at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
        at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:388)
        at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:243)
[...]

PR #2532, opened 2018-10-05
Patch macro names that start with number

kasemir:
See #2531


kasemir:
Will, this is based on an issue with the ADL converter. You care more about the EDM converter, but maybe you have an opinion anyway if it's reasonable to patch numeric macro names to allow the ADL conversion, or not, i.e. ADL files need to be fixed before conversion.

kasemir:
User who reported the problem decided to insteafd fix the *.adl file.
From #2531: "....  mca_ROIS.adl related display expected macro names N1, N2, ... but it was being called incorrectly with 1, 2, .... So I just needed to fix the adl file that was calling mca_ROIs.adl"

willrogers:
I think that probably makes sense.

PR #2533, opened 2018-10-10
Changes to allow building using JDK11.

willrogers:
These changes build and run the basic functionality. There are also changes to `maven-osgi-bundles` and `diirt` required. Don't merge, this is for discussion.

The biggest problem is JAXB - I've had to hard-code to the Eclipse Moxy implementation when it should select dynamically according to the plugins available.


berryma4:
#2530

berryma4:
closing

PR #2538, opened 2018-11-01
phoebus-integration

claudio-rosati:
Fixed missing plugin and module.

@shroffk 
I can't having the plug-ins working. I don't remember if there is a setting or a convention on where Java 11 is and where Phoebus applications are.

Should we add a preference pane?


shroffk:
I think there is a preference file already.

claudio-rosati:
Yes, I've found it. Thank you

```
org.csstudio.phoebus.integration/phoebus_home=<path-to-phoebus-repo>/phoebus-product/target
org.csstudio.phoebus.integration/jdk9_home=<path-to-jdk9/10/11>
```

Issue #2553, opened 2018-12-19
Archive Engine: Nanoseconds Incorrect


DominicOram:
At ISIS we've had some issues with the timestamp of the data being written by the archive engine:

![image](https://user-images.githubusercontent.com/10086797/50219819-4d5eba00-0388-11e9-9c13-1b6f3ce3a102.png)

Specifically it looks like the nanosecond field is being written to incorrectly (for more details see https://github.com/ISISComputingGroup/IBEX/issues/3779). We are using Archive Engine 4.0.0.201806141728 and mySQL 5.7

@kasemir have you seen this issue before?

kasemir:
Nadine at ITER had noticed this.
The Java Timestamp has nanosecond accuracy.

With Oracle, we simply use JDBC `statement.setTimestamp(..)` to set the full timestamp because the TIMESTAMP column in the database can handle it.

With Postgres and MySQL, the TIMESTAMP column is only good to seconds, and we set the nanoseconds to a separate nanosecs column. Nadine noticed, however, that something rounds the seconds, so we need to truncate the seconds before writing them.

On 2018-10-19 she emailed this diff that she had tested:

```
    private void completeAndBatchInsert(
            final PreparedStatement insert_xx, final RDBWriteChannel channel,
            final Timestamp stamp, final int severity,
            final Status status) throws Exception
    {
        // Set the stuff that's common to each type
        insert_xx.setInt(1, channel.getId());
        insert_xx.setTimestamp(2, stamp);
        insert_xx.setInt(3, severity);
        insert_xx.setInt(4, status.getId());
-       // MySQL nanosecs
-       if (rdb.getDialect() == Dialect.MySQL  ||  rdb.getDialect() == Dialect.PostgreSQL)
 
+       if (rdb.getDialect() == Dialect.Oracle)
+            insert_xx.setTimestamp(2, stamp);
+        else
+        {
+            // Truncate the time stamp
+            Timestamp truncated = Timestamp.from(stamp.toInstant());
+            truncated.setNanos(0);
+            insert_xx.setTimestamp(2, truncated);
+            // Set the nanos in a separate column
            insert_xx.setInt(6, stamp.getNanos());
        }
```

DominicOram:
Great, thanks. I'll test that my end and, assuming it works, make a PR for it.

kasemir:
[...]

PR #2554, opened 2018-12-19
#2553 Archive Engine: Nanoseconds Incorrect

herreri:
Timestamp truncated for MySQL and PostgreSQL to avoid rounding it "up" to the nearest second


kasemir:
Fixes #2553

Issue #2503, opened 2018-08-23
BOY Intensity chart interpolation


kasemir:
Interpolation settings of the BOY intensity chart were left to the Eclipse SWT/Draw2d internals.

An image that used to look like this, showing individual detector pixels, ..:

![bl14b_2d_sharp](https://user-images.githubusercontent.com/1932421/44551274-a8431600-a6f4-11e8-8f59-e83ffaa54a5a.png)

.. today looks like this, because it's auto-interpolated:

![bl14b_2d_blurry](https://user-images.githubusercontent.com/1932421/44551297-b8f38c00-a6f4-11e8-8e2f-3e52bfb95f99.png)

Need to control the interpolation settings.

kasemir:
Fix is to set interpolation based on image vs. widget size.

If widget is larger than raw image data, turn interpolation off to show detector pixels:

![fix](https://user-images.githubusercontent.com/1932421/44551335-e0e2ef80-a6f4-11e8-98fa-f32191014c98.png)

PR #2559, opened 2019-02-05
IntensityGraph: Turn interpolation on/off based on data vs. wdiget size

berryma4:
#2503
cherry pick to 4.5.x


berryma4:
@willrogers this is just timing out, could you merge :)
The time out is from all the different versions in the 4.5.x p2 repo.

willrogers:
Looks fine to me.

Issue #2579, opened 2019-04-23
Contribute Elasticsearch Messagelog history


MichaelRitzert:
I'd like to contribute the Elasticsearch based logging solution that we have been using for a while now. There are a few parts building on each other:
– a very simple interface to Elasticsearch (via JSON queries and HTTP)
– a model merging initial data retrieved from a database with live data coming in via JMS + an implementation using ES as the database. From the design of the code, it should also be possible to include PostgreSQL as another option.
– the corresponding GUI forked from org.csstudio.alarm.beast.msghist. This also includes an OPI widget.

Currently, I'm using org.json and com.sun.jersey, so this targets Java 8 and the CSS 4.5.x branch. For Java 11+ / CSS master, switching the the respective standard solutions (javax.json and java.net.http) is prepared.

If there is interest, I'd prepare a pull request to have it in the main release. Any proposal where such code could go?

Additionally, there's the C++ library that we use to perform logging via JMS from the IOCs. Most likely this will just go on our github page as its own project.

shroffk:
Hello,
You can create a new module `org.csstudio.logging.es` under the core/utility module. 

I am not sure about the C++ libraries location.

MichaelRitzert:
OK. I will create a pull request for the actual code.
We will then have to see about how to integrate it in the build, i.e. features and repo.

MichaelRitzert:
I'm depending on too many things from outside core, so I don't think it can go there. There are at least dependencies on org.csstudio.apputil for StartEndDialog, and additionally org.csstudio.opibuilder for the OPI widget.

shroffk:
ok, you are right...if it depends on non core things then it might be best to move it to applications

shroffk:
I think diag or appunorganized would both work

MichaelRitzert:
I think I have the build working with the ElasticSearch client in utilities and the rest of the code in appunorganized. The remaining question is about the features. The ES client is tiny, I just added this to core.utility.feature. The ES client is a bit larger. Shall I create a new feature for it?

berryma4:
Yes, please create a new feature.

MichaelRitzert:
I updated the pull request accordingly. #2583

MichaelRitzert:
So the pull request has been merged, but now the build again times out
https://travis-ci.com/ControlSystemStudio/cs-studio/builds/123782993 so that the new features are not yet available in the p2 repo.
How can we proceed?

[...]

PR #2583, opened 2019-04-24
Import message log based on Elasticsearch.

MichaelRitzert:
#2579 
I folded the view, model and widget into one plugin, but kept the ES client separate since I require this also in other code.


MichaelRitzert:
It seems there is a problem with the dependencies on org.csstudio.apputil.

MichaelRitzert:
Can I please get a review for this?

kasemir:
Sorry, haven't been involved in this, but as far as I can tell this is about sending the log messages to ES and having a message viewer for it. Shouldn't affect those who don't want to use it.
But there's the build errors, and since this PR was created, the build settings have been changed in the process of moving to Java 9/10/11/12. So before merging it, I think we need to get the build to pass.
Last error I saw was basically `The job exceeded the maximum time limit for jobs, and has been terminated.`, so I've started the build again.

MichaelRitzert:
Still no luck, it seems.

Note: I will create another pull request to master which uses java.net.http and javax.json (which is probably a problem now…), once the build is stable for master. But I will also need this version of code in 4.5.

MichaelRitzert:
Is there anything I can do to get this accepted? The last "error" was "The job exceeded the maximum time limit for jobs, and has been terminated."

berryma4:
I can look at this on Monday for merging.

MichaelRitzert:
Thank you. With the structure of the code settled, I will prepare the corresponding request for master.

Issue #2589, opened 2019-04-29
cs-studio project doesn't build with Java 11.0.2


claudio-rosati:
Hello,

my Java 11 pipeline fails at this point:

```
[2019-04-29T01:47:54.100Z] Running org.csstudio.logbook.olog.property.fault.FaultConfigurationParsingTest
[2019-04-29T01:47:54.100Z] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.041 s <<< FAILURE! - in org.csstudio.logbook.olog.property.fault.FaultConfigurationParsingTest
[2019-04-29T01:47:54.100Z] parserTest(org.csstudio.logbook.olog.property.fault.FaultConfigurationParsingTest)  Time elapsed: 0.018 s  <<< ERROR!
[2019-04-29T01:47:54.100Z] javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
[2019-04-29T01:47:54.100Z] 	at org.csstudio.logbook.olog.property.fault.FaultConfigurationParsingTest.parserTest(FaultConfigurationParsingTest.java:24)
[2019-04-29T01:47:54.100Z] Caused by: java.lang.ClassNotFoundException: com.sun.xml.internal.bind.v2.ContextFactory
[2019-04-29T01:47:54.100Z] 	at org.csstudio.logbook.olog.property.fault.FaultConfigurationParsingTest.parserTest(FaultConfigurationParsingTest.java:24)
[2019-04-29T01:47:54.100Z] 
[2019-04-29T01:47:54.100Z] Running org.csstudio.logbook.olog.property.fault.FaultAdapterTest
[2019-04-29T01:47:54.100Z] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s - in org.csstudio.logbook.olog.property.fault.FaultAdapterTest
[2019-04-29T01:47:54.100Z] checkCauseCorrectionFaultTextParsin(org.csstudio.logbook.olog.property.fault.FaultAdapterTest)  Time elapsed: 0.001 s
[2019-04-29T01:47:54.100Z] checkDescRepairFaultTextParsin(org.csstudio.logbook.olog.property.fault.FaultAdapterTest)  Time elapsed: 0 s
[...]

PR #2598, opened 2019-06-12
Attempt to fix tests that use JAXB after JDK-11 changes.

willrogers:
These copy the changes in the rest of cs-studio, but the test still
fails to find JAXB-API.

Don't merge! This doesn't fix the problem for me.

@berryma4 @shroffk

### DIFF
--- a/applications/logbook/logbook-plugins/org.csstudio.logbook.olog.property.fault.test/META-INF/MANIFEST.MF
+++ b/applications/logbook/logbook-plugins/org.csstudio.logbook.olog.property.fault.test/META-INF/MANIFEST.MF
@@ -6,4 +6,9 @@ Bundle-Version: 1.0.0.qualifier
 Fragment-Host: org.csstudio.logbook.olog.property.fault
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Bundle: org.junit;bundle-version="4.11.0",
- org.hamcrest.core;bundle-version="1.1.0"
+ org.hamcrest.core;bundle-version="1.1.0",
+ jaxb-api;bundle-version="2.3.1",
+ com.sun.xml.bind.jaxb-impl;bundle-version="2.3.1",
+ javax.activation;bundle-version="1.1.1",
+ javax.activation-api;bundle-version="1.2.0"
+Import-Package: com.sun.xml.bind.v2
--- a/applications/logbook/logbook-plugins/org.csstudio.logbook.olog.property.fault.test/src/org/csstudio/logbook/olog/property/fault/FaultConfigurationParsingTest.java
+++ b/applications/logbook/logbook-plugins/org.csstudio.logbook.olog.property.fault.test/src/org/csstudio/logbook/olog/property/fault/FaultConfigurationParsingTest.java
@@ -14,23 +14,32 @@ import javax.xml.bind.Unmarshaller;
 import javax.xml.transform.stream.StreamSource;
 
 import org.csstudio.logbook.olog.property.fault.FaultConfiguration.Group;
+import org.csstudio.logbook.olog.property.fault.FaultConfigurationFactory;
 import org.junit.Test;
 
 public class FaultConfigurationParsingTest {
 
     @Test
     public void parserTest() throws JAXBException {
+        final ClassLoader orig = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(JAXBContext.class.getClassLoader());
 
-        JAXBContext context = JAXBContext.newInstance(FaultConfiguration.class);
-        Unmarshaller um = context.createUnmarshaller();
-        JAXBElement<FaultConfiguration> fc2 = um.unmarshal(new StreamSource(new File("resources/default_fault_config.xml")),
-                FaultConfiguration.class);
-        FaultConfiguration fc = fc2.getValue();
-        assertEquals("areas not equal", fc.getAreas(), Arrays.asList("Global", "Linac", "BR", "SR"));
-        assertEquals("subsystems not equal", fc.getSubsystems(), Arrays.asList("Diagnostics", "EPS", "RF", "Controls"));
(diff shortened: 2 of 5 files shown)


georgweiss:
I see the same here at ESS.

willrogers:
I don't know why this doesn't work, @berryma4 is the most likely to understand. Is the test classloader different to the runtime classloader?

georgweiss:
What about considering a different XML parsing strategy? Like XMLBeans or DOM.

kasemir:
DOM's what I tend to use...

berryma4:
Sorry, I was on vacation for a bit ...
I'm going to assume the issue is that you need to pass the JAXBContextFactory parameter through mvn:
```
-Djavax.xml.bind.JAXBContextFactory=com.sun.xml.bind.v2.ContextFactory
```

georgweiss:
I'm sorry, but -Djavax.xml.bind.JAXBContextFactory=com.sun.xml.bind.v2.ContextFactory does not make a difference for the build on my box. Still fails in the same manner.

willrogers:
It looks like it should be `javax.xml.bind.context.factory`, but that still doesn't work for me in Tycho.

berryma4:
Your right, it seems the property is not read reliably in maven. I added a jaxb.properties file, which might be better anyway.

willrogers:
I also needed a `jaxb.index` file. After that the test passed.

I'm not sure how you pushed to this branch.

berryma4:
I was surprised I could push too!

berryma4:
It might be a good idea to not use the environment variable, and drop the property files in everywhere needed instead.

berryma4:
I wonder why I didn't need the ```jaxb.index``` file to pass the test? It's more correct to have it though.

willrogers:
I also had to replace some tabs with spaces, but it builds for me now. I'm happy to merge.

berryma4:
#2589

georgweiss:
Build now fails at later point:

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.004 s - in org.csstudio.swt.rtplot.TraceAnalyzerTest
testFindSample(org.csstudio.swt.rtplot.TraceAnalyzerTest)  Time elapsed: 0.003 s
Running org.csstudio.swt.rtplot.LinearTicksTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s - in org.csstudio.swt.rtplot.LinearTicksTest
testNiceDistance(org.csstudio.swt.rtplot.LinearTicksTest)  Time elapsed: 0.002 s
testPrecision(org.csstudio.swt.rtplot.LinearTicksTest)  Time elapsed: 0 s

Results:

Failures: 
  UpdateThrottleTest.testUpdateThrottle:63 
Expected: <2>
     but: was <3>

Tests run: 13, Failures: 1, Errors: 0, Skipped: 0

kasemir:
`UpdateThrottleTest` could be sensitive to timing and CPU load.
Does it fail each time?

willrogers:
It's passing for me.

kasemir:
We have several tests which prove that Java is not a real-time system. They do occasionally fail.

georgweiss:
One might argue that the test in question should be reworked.

kasemir:
Yes, ideally by somebody who sees it fail right now, so you can adjust the timing.

kasemir:
For that specific one, we've tried several times when we hit timing constraints:
```
git log UpdateThrottleTest.java
commit 975513d5051fb958ef7e9cca1fd6594a95e931dc
Author: Claudio Rosati <Claudio.Rosati@esss.se>
Date:   Thu Jul 6 15:44:13 2017 +0200
    o.c.swt.rtplot.test: update tests
commit 73499d9b94333aca82bfecdc5034de4a9d94706a
Author: kasemir <kay.kasemir@gmail.com>
Date:   Mon Nov 7 16:27:31 2016 -0500
    o.c.swt.rtplot.test: Increase times in time-related test
    Should help avoid test failures as seen in #1947
commit f3b21f12dc45be3654c43da4f52d35e5d8b00e45
Author: kasemir <kasemirk@ornl.gov>
Date:   Thu Sep 15 10:14:40 2016 -0400
    swt.rtplot: Increase time delay to allow test to pass
    #1898
```

kasemir:
It's easy to throw stones at this type of test, but I'm not sure how to improve them.
This is about testing time-related logic, like a throttle which suppresses some events but then issues an event, ideally within some time.
So we expect an event within say half a second.
Just to make sure we wait 2 seconds.
Most of the time this works out, but especially on cloud setups (travis, ..) it often takes a lot longer.
What to do? Wait 60 seconds? That defeats the test.
Skip the test because timing can't be dependable?
For now I prefer to have the test and just ignore occasional failures, because most of the time it works and might help to catch a real problem if it suddenly fails every time.

georgweiss:
What about rewriting the test to wait for 2 seconds, and exponentially increase the wait time up to a limit where we consider the test useless.

kasemir:
Well, that's a little nicer than this specific test, but doesn't solve the failures.
Assume we expect something to happen in 0.5 seconds.
Then this is clearly a waste of time because you always delay for 5 seconds:
```
TimeUnits.SECONDS.sleep(5);
assert(.. that it happened);
```

Instead of inventing a new machinery that sleeps 1 second, then 2, then 4, .. newer tests tend to use for example a `CountDownLatch` or a `BlockingQueue` so you can wait for the result like this:
```
assert(  block.poll(5, TimeUnit.SECONDS) == expected_result );
```

That way we don't waste time waiting; you move on as soon as you have the result, or time out.

The big issue, however, is that these tests sometimes do time out when run on build servers.
Take https://github.com/shroffk/phoebus/blob/master/services/alarm-server/src/test/java/org/phoebus/applications/alarm/server/AutomatedActionTest.java#L217:
It usually runs fine, reacting within 1 ms of the expected time:
```
Action performed after 1001 ms (expected: 1000 ms)
```

But I just got a travis email that it timed out after waiting for 5 seconds.
You'd think that 5 seconds is ample time for what usually happens in 1ms.

Issue #2584, opened 2019-04-25
SVG batik warning on /BOY Examples/pictures/SymbolLibrary/SVG/PL General.svg


berryma4:
```java
2019-04-25 09:53:59.766 WARNING [Thread 1] org.csstudio.swt.widgets.symbol.SVGSymbolImage (loadDocument) - Error loading SVG image /BOY Examples/pictures/SymbolLibrary/SVG/PL General.svg
org.apache.batik.dom.util.SAXIOException: The prefix "xlink" for attribute "xlink:href" associated with an element type "use" is not bound.
	at org.apache.batik.dom.util.SAXDocumentFactory.createDocument(SAXDocumentFactory.java:459)
	at org.apache.batik.dom.util.SAXDocumentFactory.createDocument(SAXDocumentFactory.java:357)
	at org.apache.batik.anim.dom.SAXSVGDocumentFactory.createDocument(SAXSVGDocumentFactory.java:226)
	at org.csstudio.swt.widgets.symbol.SVGSymbolImage.loadDocument(SVGSymbolImage.java:312)
	at org.csstudio.swt.widgets.symbol.SVGSymbolImage$2.runWithInputStream(SVGSymbolImage.java:280)
	at org.csstudio.swt.widgets.util.AbstractInputStreamRunnable.run(AbstractInputStreamRunnable.java:32)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:185)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4906)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4475)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1173)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1062)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:156)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:636)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
[...]

PR #2601, opened 2019-06-28
Split of Batik bundles

GuillemCastro:
As a consequence of https://github.com/ControlSystemStudio/maven-osgi-bundles/pull/85 we had to update some manifests to work with Batik subjars.

This should solve https://github.com/ControlSystemStudio/cs-studio/issues/2448

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.editor/META-INF/MANIFEST.MF
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.editor/META-INF/MANIFEST.MF
@@ -21,7 +21,8 @@ Require-Bundle: org.eclipse.core.filesystem;bundle-version="1.2.0",
  org.jdom,
  org.eclipse.ui.workbench;bundle-version="3.106.0",
  org.eclipse.ui.views;bundle-version="3.7.0",
- org.eclipse.ui
+ org.eclipse.ui,
+ org.apache.batik.util;bundle-version="[1.10.0,1.11.0)"
 Bundle-Vendor: Xihui Chen<chenx1@ornl.gov> - SNS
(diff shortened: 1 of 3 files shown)


berryma4:
Great news!

berryma4:
#2584

berryma4:
@GuillemCastro does this work running the product from the IDE?

berryma4:
Adding @shroffk to the conversation.

shroffk:
The current master version of org.csstudio.utility.batik MANIFEST files complains

image

GuillemCastro:
I added this "Export-Package" because otherwise the compilation from the command line would fail. I agree that it is not correct because the plugin doesn't include the `org.apache.batik.util` package. The bundle that exports this package is available from the ORBIT site but it seems like Maven/Tycho weren't able to find it.

I've found a workaround. I'm preparing a PR.

Also, while looking at this issue, I've found another. Eclipse 2019-03 reports several errors like,

`The package org.w3c.dom is accessible from more then one module: <unnamed>, java.xml`
`The package javax.xml.parsers is accessible from more then one module: <unnamed>, java.xml`

For this reason, some the imports of classes in these packages also fail.

The `org.csstudio.utility.batik` bundle has a dependency on `javax.xml` (that exports the `javax.xml` package) which will be added by Eclipse to the classpath (which is added to the unnamed module). A module of the JDK already exports the `javax.xml` package, and according to the Java specification this is ilegal (see https://stackoverflow.com/questions/51094274/eclipse-cant-find-xml-related-classes-after-switching-build-path-to-jdk-10). You cannot have a package split between the classpath and the modulepath.

It seems like `javac` has a bug (https://bugs.openjdk.java.net/browse/JDK-8215739) and is not enforcing this rule, but Eclipse (since 2019-03) is. It _should_ be fixed in JDK 14.

Issue #2607, opened 2019-07-24
Phoebus find open port and app in path


berryma4:
I would like to have the integration plugin pick an open port, and find phoebus in the user path.

shroffk:
Using the random port seems to not make much sense to me.
In that case it might be best to not use the ```-server``` options, so that each launch is a standalone instance of phoebus

berryma4:
I was thinking one server instance (with port) per eclipse instance.

shroffk:
I see...
Can I implement the following:
```
# launch the phoebus application with the -server argument, use a random available port
use_phoebus_server=true
```

This way, if my launcher script takes care of the port number then I can disable that by setting the above property to false

berryma4:
My pull request you merged works. But, I'm combining two things, find
Phoebus in the user's path and pick an open port.

shroffk:
I know, I have added on top of that

shroffk:
https://github.com/ControlSystemStudio/cs-studio/pull/2610

berryma4:
Thank you!

PR #2608, opened 2019-07-24
o.c.phoebus.integration: find empty port, find phoebus in user path

berryma4:
#2607

### DIFF
--- a/applications/phoebus/phoebus-plugins/org.csstudio.phoebus.integration/src/org/csstudio/phoebus/integration/PhoebusLauncherService.java
+++ b/applications/phoebus/phoebus-plugins/org.csstudio.phoebus.integration/src/org/csstudio/phoebus/integration/PhoebusLauncherService.java
@@ -5,9 +5,11 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.ServerSocket;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


berryma4:
I think we could probably organize this set of plugins different. But, to start I wanted to loosen the coupling a bit (ie. needing a version, port number, etc.)

shroffk:
I have a set of changes that I need to make for the phoebus integration...
however I am stuck with the previously mentions issues with running cs-studio in eclipse.

berryma4:
I see the same issue (batik). It looks like a split package resolving issue.

shroffk:
Is everyone able to successfully run/debug cs-studio from inside cs-studio?

kasemir:
I like the idea of having the integration plugin pick a port, so each Eclipse instance interfaces with its own phoebus instance.

But related to the eclipse issues: All the recent pull requests were merged with failing travis builds. Can anybody build and run this stuff?

berryma4:
The travis builds are timing out. If we email they usually extend it. It does take longer to build with this version of tycho, it is fix in 1.4, but it has other issues ....

kasemir:
My eclipse setup is based on sources from March, the "java8" branch.
For some reason, maven doesn't build with the java8 branch.
So since March I have no local CS-Studio updates because both building the old branch with Java 8 and working on the 'master' with Java 9/10/11/12 is broken.

shroffk:
I can build it on my CI server
I can build it on the command line on my machine

I can't get it to run properly inside the IDE...
https://github.com/ControlSystemStudio/cs-studio/pull/2601#issuecomment-512957155

I know this is specifically an IDE issue cause the product I build runs fine. It resolves batik and the rest of the dependency chain fine.

However, not having things working in the IDE really makes it difficult to develop... after 2 days of trying to understand why eclipse **alone** does not like how the dependencies are sorted I switched to another task.

kasemir:
Why are the travis builds still failing when the maven build otherwise works for you?

shroffk:
like eric said...this has something to do with the travis infrastructure/resource and build duration

shroffk:
I have actually not had much trouble getting it to build...but getting it to properly run inside the IDE for development purposes is something I still can't get to work.

berryma4:
@shroffk you should be okay now, after #2609

berryma4:
@shroffk  could we merge this?

berryma4:
thank you!

PR #2609, opened 2019-07-25
Removed Import-Package that fails in Eclipse

GuillemCastro:
This PR solves the issue introduced with https://github.com/ControlSystemStudio/cs-studio/pull/2601. With this commit it should be possible to use CSS from Eclipse (up to 2018-12) again.

Note that in Eclipse 2019-03 and later, it won't work as described in https://github.com/ControlSystemStudio/cs-studio/pull/2601#issuecomment-514946448


berryma4:
Thank you!

PR #2620, opened 2019-11-20
removing the import-packages in preference for required bundles

shroffk:


shroffk:
@kasemir or @dylan171 could you please review/approve this

PR #2622, opened 2019-12-17
More istack patching.

willrogers:
@shroffk we were still getting a failing test in our Jenkins build.

I copied your recent changes and it appears to have fixed the problem.

cc @ifmalcuaz


shroffk:
good... I think we should move away from the import packages for the sake of our sanity

PR #2626, opened 2020-01-27
Allow the launching of phoebus without server

shroffk:
At nsls2, the run-phoebus script takes care of finding the port so the phoebus launcher service only needs to launch phoebus using the system launcher and not worry about the port.


shroffk:
@willrogers could you please review this... thanks!!

PR #2633, opened 2020-02-11
2618 restructured repo

shroffk:
This is in preparation for the newly structured repo


shroffk:
To test the build:
checkout the 2618 branch

mvn clean verify -D eclipse.p2.mirrors=false -D org.osgi.framework.bundle.parent=ext -D osgi.framework.extensions=org.eclipse.fx.osgi -D baselineMode=warn -D jgit.dirtyWorkingTree=warning -D tycho.localArtifacts=ignore -D csstudio.composite.repo=C:\git\cs-studio\p2repo-4.6 -D skipTests=true

shroffk:
NOTE:
for some people we had to install the maven-osgi-bundles once first because tycho failed to resolve the dependencies that were being installed by it.

DominicOram:
There's a script for rewriting history at https://gist.github.com/emiller/6769886

willrogers:
The history of all parts of the repo has been retained, but since the files have have been moved there is a short history for each file.

You can see the entire history of a file using e.g. `git log --follow maven-osgi-bundles/pom.xml`.

To me this seems sufficient and probably preferable to trying to rewrite history.

DominicOram:
> You can see the entire history of a file using e.g. `git log --follow maven-osgi-bundles/pom.xml`.

Both github and eclipse don't do this `follow` so history appears lost in them.

I'm on the fence about re-writing the history.

DominicOram:
We can run mvn multithreaded using `-T`: https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3

willrogers:
I can make this work with the following three steps:

* `mvn clean verify` from the root - this fails on the diirt datasource-extra problem
* `mvn clean verify` inside the `core/diirt` directory - this succeeds
* `mvn clean verify` from the root again - now this succeeds

willrogers:
A cleaner option: remove `maven-osgi-bundles` from the main build. It doesn't seem to be straightforward to resolve the entire maven project when some of the bundles don't exist until `maven-osgi-bundles` has been built.

If you remove `maven-osgi-bundles` then you don't have this resolution problem. However, you have to build it in a separate step.

DominicOram:
> A cleaner option: remove `maven-osgi-bundles` from the main build. It doesn't seem to be straightforward to resolve the entire maven project when some of the bundles don't exist until `maven-osgi-bundles` has been built.
> 
> If you remove `maven-osgi-bundles` then you don't have this resolution problem. However, you have to build it in a separate step.

This worked for me too

PR #2642, opened 2020-05-11
Numeric to string for alarm status

tanviash:
Currently, Databrowser "Inspect samples" view shows the alarm status as a numeric enum value when source is archiver appliance. This PR is to provide numeric to string conversion for alarm status field so that status field is represented as string enum as it is more user friendly.


applications/archive/archive-plugins/org.csstudio.archive.reader.appliance/src/org/csstudio/archive/reader/appliance/ApplianceValueIterator.java
```diff
@@ -320,4 +322,15 @@ protected static AlarmSeverity getSeverity(int severity) {
             return AlarmSeverity.UNDEFINED;
         }
     }
+
+    /**
+     * Determines alarm status from the given numerical representation.
+     *
+     * @param status numerical representation of alarm severity
```
kasemir, line 329: .. numerical representation of alarm _status_, not severity

tanviash: Corrected

dylan171:
@tanviash For testing at FRIB I left off the version from the manifest, so I simply used `org.epics.jca,`

willrogers:
It looks to me like some of the tests fail because of this change. It's not your fault, we should be running them!

Issue #2640, opened 2020-04-15
Another Batik dependency problem


willrogers:
I have built the latest master branch. I then run the following command:

```
./cs-studio -nosplash -application org.csstudio.opibuilder.converter.edl -h
```
I get a little pop-up saying 'An error has occurred. Please see the log file ...

I have attached the log file. Unfortunately it is another batik depdendency error and I cannot figure it out. Any ideas @shroffk?

I first saw this when re-enabling the unit tests on our Jenkins build.

[1586968616661.log](https://github.com/ControlSystemStudio/cs-studio/files/4482521/1586968616661.log)

shroffk:
Hmm... No I have not come across this one as yet.
I don't use the converter much which is why we might have missed it.

willrogers:
I think all you need to do to find this error is run the unit tests.

willrogers:
I think this was resolved by #2647 .

PR #2643, opened 2020-07-01
Update .travis.yml

shroffk:


willrogers:
Aha @shroffk I think you have the same problem I described in #2640. Do you think it is the same thing?

My manager would now like me to try to resolve this build problem. I would rather eat slugs but there we are.

willrogers:
I have a 'fix' for the Batik problem that allows the build to succeed. Whether it works correctly at runtime is not certain.

This should now be failing on genuine test failures introduced in #2642.

shroffk:
Ok,
Should this be merged?

shroffk:
I was able to recreate the test failures and will look into them tomorrow

willrogers:
I did indeed get a runtime problem. The latest changes do appear to work on the one example of SVGs that I can find, although these widgets are a bit fiddly.

I also fixed the tests.

shroffk:
I know that ITER uses SVGs a lot, can someone from @css-iter test this?

shroffk:
@willrogers can you add your review please

Issue #2638, opened 2020-02-13
Create Log Entry Dialog Failure


dxmaxwell:
When attempting to create a log entry with an attachment, the following error occurs:
![CS Studio Create Log Entry Dialog with Error](https://user-images.githubusercontent.com/109703/74457736-ad4ed480-4e56-11ea-8a3b-bb96b25ae52f.png)
![CS Studio Create Log Entry Error Dialog](https://user-images.githubusercontent.com/109703/74457747-b17af200-4e56-11ea-8b93-6bd034b0ca04.png)

The stack trace is as follows:
```
javax.ws.rs.WebApplicationException: javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: com.sun.xml.bind.v2.ContextFactory]
	at com.sun.jersey.core.provider.jaxb.AbstractRootElementProvider.readFrom(AbstractRootElementProvider.java:115)
	at com.sun.jersey.api.client.ClientResponse.getEntity(ClientResponse.java:634)
	at com.sun.jersey.api.client.ClientResponse.getEntity(ClientResponse.java:586)
	at com.sun.jersey.api.client.WebResource.handle(WebResource.java:686)
	at com.sun.jersey.api.client.WebResource.access$200(WebResource.java:74)
	at com.sun.jersey.api.client.WebResource$Builder.post(WebResource.java:570)
	at edu.msu.nscl.olog.api.OlogClientImpl.add(OlogClientImpl.java:831)
	at org.csstudio.logbook.olog.OlogLogbookClient.addAttachment(OlogLogbookClient.java:203)
	at org.csstudio.logbook.olog.OlogLogbookClient.createLogEntry(OlogLogbookClient.java:162)
	at org.csstudio.logbook.ui.LogEntryBuilderDialog$1.run(LogEntryBuilderDialog.java:144)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: com.sun.xml.bind.v2.ContextFactory]
	at javax.xml.bind.ContextFinder.newInstance(ContextFinder.java:278)
	at javax.xml.bind.ContextFinder.find(ContextFinder.java:397)
[...]

PR #2646, opened 2020-07-28
#2638 fixing the javax warning while adding attachments

shroffk:
@dxmaxwell moving fix from the RIP repo to the main cs-studio one


shroffk:
@dxmaxwell can you approve this PR

PR #2647, opened 2020-09-02
Update preparation

willrogers:
@shroffk these changes are necessary to build against Eclipse 2020-06. As before, I need to check runtime behaviour as well as the build passing.

I haven't made that update but once these changes are merged I am able to build against that version using Maven properties.


shroffk:
Ok, 
The updating of the tycho version makes me a bit nervous.

When do you want to merge this?

willrogers:
Tycho hasn't been difficult recently.

Wait until I've made sure everything is working please.

willrogers:
I think these changes are fine to be merged.

I also think we should start a 4.7 branch - I've now got a build that works against 2020-09 but that will want to be updated to 2020-12 at some point.

PR #2657, opened 2020-09-03
Pomless builds

vogella:
This enableds pomless builds. For testing the pom from org.cssstudio.startup.intro has been removed. Other plug-in specific pom files files could also be deleted


willrogers:
Included in #2658.

PR #2662, opened 2020-12-09
Css392 enum labels

rjwills28:
Adding the ability to display the enum labels (and not just the value) in the databrowser 'Cursor Value' field and in the Cross-hair tool for an enum type PV.
E.g. Cursor value: Open (1) instead of simply <1>.


willrogers:
@kasemir @shroffk this fills in the enum labels from the archiver appliance, which our users have been missing.

I am not sure how general extracting the enum label for the crosshair will be.

shroffk (approved):
Looks fine to me

PR #2663, opened 2020-12-09
Css546 stddev plot

rjwills28:
It has been discussed and agreed that the standard deviation plotted in the databrowser is confusing and not that useful (see https://github.com/ControlSystemStudio/phoebus/issues/1656). The plotting of the standard deviation has therefore been removed.

In addition, the auto-scaling of the y-axis does include the min and max values plotted in the AREA trace. The min and max can often fall outside of the y-axis range and so are not always easily visible. This set of changes instead uses the min/max (where applicable) to determine the the appropriate y-axis range so that they are always visible.

### DIFF
--- a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
+++ b/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
@@ -84,12 +84,32 @@ public class PlotProcessor<XTYPE extends Comparable<XTYPE>>
                         {
                             final PlotDataItem<XTYPE> item = data.get(i);
                             final double value = item.getValue();
-                            if (! Double.isFinite(value))
-                                continue;
-                            if (value < low)
-                                low = value;
-                            if (value > high)
-                                high = value;
+                            final double max = item.getMax();
+                            final double min = item.getMin();
+                            
+                            if (!Double.isFinite(max)) 
+                            {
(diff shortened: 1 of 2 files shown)


applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
```diff
@@ -84,12 +84,32 @@ public ValueRange call() throws Exception
...
                             final double value = item.getValue();
-                            if (! Double.isFinite(value))
-                                continue;
-                            if (value < low)
-                                low = value;
-                            if (value > high)
-                                high = value;
+                            final double max = item.getMax();
+                            final double min = item.getMin();
+
```
willrogers, line 89: We could do with a comment explaining what's going on here.

willrogers:
@kasemir @shroffk these changes just make the plots clearer.

applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/PlotProcessor.java
rjwills28: I have just added comments to the code to describe how the min, max or value are used to determine the range. Please see latest [commit](https://github.com/ControlSystemStudio/cs-studio/pull/2663/commits/67d15b079ae937c455c507015efcb9cd7a52d901).

kasemir (approved):
Fine with me

PR #2664, opened 2021-01-07
Ci updates

willrogers:
@shroffk I don't think these are controversial, but let's make sure they run. I think we will have to remove Travis soon.


willrogers:
Yes, we have run out of Travis credits so I have removed `.travis.yml`.

shroffk (approved):

PR #2668, opened 2021-01-08
Fixing bug that causes archived data not to be plotted after it has been fetched

rjwills28:
We found an issue where the plot is not updated immediately after data is fetched from the Archiver Appliance. Instead, in order to get the new data plotted, the user has to interact with the plot in some way which will call an update to the plot and so show the new data. 

The issue is that the databrowser is not registering that there are new samples for a PV item. This is because when the new samples are merged with the current values, the `have_new_samples` boolean is not set to true (as is done in the case of live data). A value of true is required in order for a redraw to be executed.


kasemir (approved):
Indeed. Same was done to databrowser3 in Aug 2017, https://github.com/kasemir/org.csstudio.display.builder/commit/bf4d6f4440b37ba1bbb2c354ffb2b48f3ad6a055

PR #2669, opened 2021-01-10
New preference to configure what CS-Studio should do when closing the last window.

rjwills28:
Gnome3 does not have a tray and so the option that is displayed when the last CS-Studio window is closed to minimize to tray makes no sense in this case. If the window is minimized to tray it then becomes impossible to get back. 

We have created a new preference that dictates what to do when closing the last CS-Studio window. There are 3 options: ask to minimize to tray (which will bring up the previous minimize to tray dialog with the same functionality), warn that the last window is being closed (options to cancel or exit), or finally just close the window with no prompts. The 'warning of last window' dialog also allows the user to select 'do not warn again'.

### DIFF
--- a/core/utility/utility-plugins/org.csstudio.trayicon/preferences.ini
+++ b/core/utility/utility-plugins/org.csstudio.trayicon/preferences.ini
@@ -3,3 +3,4 @@
 
 minimize_to_tray = prompt
 start_minimized = false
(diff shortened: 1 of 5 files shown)


willrogers (approved):
I expect these changes only affect the DLS product.

Issue #2670, opened 2021-01-22
New feature to introduce non ignorable event on widgets and process all the elements on a queue collector. 


pedroramirezp:
I'd like to introduce these new features in order to process all the new values CSStudio receives to update a widget. We checked that there is a QueueCollector but all the values that are stored in it aren't processed, 
Moreover, not all the events that arrive at a widget are processed. these are stored in a queued with a but only one event per widget is allowed at the same time.

To get the performance described bellow is necessary:
- Modify CSStudio to process all the events a widget receives if it's configured so.
- Modify CSStudio to process all the buffered values if it's configured so.

willrogers:
Hi, sorry no-one noticed this. Have you talked to anyone in the CS-Studio collaboration before?

I didn't know that ESA was using EPICS.

dmarina-esa:
Hi. Thank you for your answer.

For what I know, we haven't talked to anyone in the CS-Studio before.
We are using some parts of the CS-Studio tools for developing User Interfaces for the last 6 years, but we don't use EPICS. 

Unfortunately, we cannot change versions easily and our baseline is Eclipse Luna for the time being. However, we will consider the use of Phoebus in the future, when the system will be migrated to a newer baseline.

Thanks again for your help and the suggestions!

kasemir:
I'm OK with the change, but some comments:

The CS-Studio UI is an operator interface, not a data acquisition system. The operator interface is meant to show the most recent value of a PV. If there are many updates, it _hopefully_ drops a few intermediated values to lessen the CPU load. That's especially true for the Eclipse-based BOY code which handles a lot of the data on the UI thread.
With the newer CS-studio version, more of the data handling is off the UI thread, but we still throttle the data for most widgets via the RXJava `throttleLatest` mechanism which reduces the update rate in a way that you get the latest value while dropping intermediates when you encounter a flurry of updates.
Only very few widgets, namely the 'Data Browser' which plots samples over time, subscribe with `BackpressureStrategy.BUFFER` to collect all received values and then periodically updating that in a plot, but note that with Channel Access the underlying network protocol will actually silently drop intermediate samples when it enters 'flow control' mode.

If you use your own network protocol, you might be able to get "all" updates into CS-Studio, and with your suggested changes you can then send "all" updates to the widgets, but beware that you might just set yourself up to GUI freezes.

willrogers:
[...]

PR #2673, opened 2021-02-02
#2670 New feature to introduce non ignorable event on widgets and ...

dmarina-esa:
New feature to introduce non ignorable event on widgets and process all the elements on a queue collector. #2670

Signed-off-by: David Marina David.Marina@esa.int


willrogers:
Hi, I just made a comment on the associated ticket.

willrogers (approved):
This change seems ok, and since it is targeted at the `4.1.x` branch it won't affect any other sites.

PR #2674, opened 2021-02-03
Configurable annotation formats and improvement of annotation editability in the databrowser

rjwills28:
This allows the user to input custom formats for the annotation values (e.g. specify the number of dp) and the timestamps. The formats are verified before adding to the plot. This is added as an 'advanced' option, while the original formatting implementation is still maintained for the general case.
In addition, two new features have been added to improve user interaction with annotations. These include being able to select 'Add annotation' from the right-click menu on the databrowser plot and being able to edit existing annotations by double-clicking on them.

### DIFF
--- a/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java
+++ b/applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java
@@ -25,6 +25,7 @@ import org.csstudio.swt.rtplot.Activator;
 import org.csstudio.swt.rtplot.Annotation;
 import org.csstudio.swt.rtplot.AxisRange;
 import org.csstudio.swt.rtplot.Messages;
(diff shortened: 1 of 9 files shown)


applications/appunorganized/appunorganized-plugins/org.csstudio.swt.rtplot/src/org/csstudio/swt/rtplot/internal/Plot.java
```diff
@@ -243,7 +246,11 @@ else if (type == Instant.class)
         () ->
         {
             plot_processor.autoscale();
-            updateImageBuffer();
+            // In Eclipse 2020-12 we cannot seem to run updateImageBuffer() on
```
willrogers, line 249: This PR shouldn't really include these lines, they're the ones regarding the plot crashing and are handled elsewhere.

rjwills28: Agreed, thanks for catching that. I have reverted this change in the latest commit.

kasemir (approved):

PR #2675, opened 2021-02-26
Optimized with last sample

willrogers:
This pull request adds a preference setting to allow using the 'optimized with last sample' algorithm that we have recently added to the Archiver Appliance: https://github.com/slacmshankar/epicsarchiverap/pull/112

The default is to use the original algorithm.

This gives much better behaviour, as discussed at some length here: https://github.com/ControlSystemStudio/cs-studio/issues/2483

cc @rjwills28


kasemir (approved):
Fine with me

PR #2676, opened 2021-02-26
Css442 opi contextmenu focus

willrogers:
After opening the context menu with right click and either selecting an item or cancelling, sometimes focus isn't returned to the OPI. This means that the next click on the OPI is not registered  as expected.

This PR fixes that problem.

cc @rjwills28


kasemir (approved):
Thanks for continuing to maintain this.

Is this a general issue, or is it specific to "SWT" vs. "Draw2D" widgets?

willrogers:
I'm not entirely sure. Becky said that the fix in the first commit worked if there was a databrowser widget (an SWT widget?) on the page but I found that it didn't always work. The second commit appears to fix the problem always.

rjwills28:
I'm not 100% sure but I feel like it might be an SWT issue from the research I was doing to find a fix. It certainly isn't an issue in Phoebus.

PR #2683, opened 2021-03-30
Css551 trace annotation visibility

willrogers:
@kasemir this is not a trivial fix, so please have a look at the implementation.


kasemir (approved):
Looks OK

rjwills28:
Something strange happened when I did a rebase with the style-checks in the newer master branch. Scrapping this PR and branch. I will create a new branch with just the correct commits and create a new PR.

kasemir:
If you're redoing it, you could also compare with the Phoebus version.
Your original approach was to add/remove traces and annotations from the plot as they are visible/hidden.
In phoebus, I now always add all traces and annotations to the plot, and simply update their 'visible' property.
That way, the 'index' of a trace in the plot always matches the 'index' in the model, overall simpler.

kasemir:
I think this one is obsolete, handled by  #2685

PR #2685, opened 2021-03-31
Trace annotation visibility (build fix)

rjwills28:
This is the same fix described in PR https://github.com/ControlSystemStudio/cs-studio/pull/2683. There was an issue with the rebase required to bring the branch up to date with the latest code style fixes as so I created a new branch and hence new PR.


rjwills28:
@kasemir I took a look at the Phoebus implementation and found a few edge cases where the annotations can get lost after hiding and showing a trace. If you hide an existing trace (with annotations), then add a brand new trace and then show the previously hidden one, it re-appears with out its annotation.
Also I noticed that the Edit Annotation window shows all of the annotations including those for the traces that are hidden.

I think both of these issues should be fixed in this PR for CS-Studio.

kasemir (approved):

PR #2690, opened 2021-06-08
Fixing bug found in the annotation visibility change that affects the…

willrogers:
… Inspect Waveform panel.

Fixing problem introduced earlier.


kasemir (approved):
Thanks!

PR #2695, opened 2021-06-16
Css534 detail panel

willrogers:
Technicality with this widget: more than the 'max' number of rows weren't handled well. Not a common case.

cc @rjwills28

### DIFF
--- a/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.detailpanel/src/org/csstudio/opibuilder/widgets/detailpanel/DetailPanelModel.java
+++ b/applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.detailpanel/src/org/csstudio/opibuilder/widgets/detailpanel/DetailPanelModel.java
@@ -38,7 +38,7 @@ public class DetailPanelModel extends AbstractContainerModel implements IPVWidge
     }
 
     /* Constants */
-    public static final int MAX_ROW_COUNT = 200;
+    public static final int MAX_ROW_COUNT = 1000;
 
     /* The widget properties */
     public static final String PROP_ROW_COUNT = "row_count";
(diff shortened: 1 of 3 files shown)


applications/opibuilder/opibuilder-plugins/org.csstudio.opibuilder.widgets.detailpanel/src/org/csstudio/opibuilder/widgets/detailpanel/DetailPanelModel.java
```diff
@@ -72,7 +72,7 @@ protected void configureProperties() {
         addPVProperty(new PVNameProperty(PROP_PVNAME, "PV Name", WidgetPropertyCategory.Basic,
                 ""), new PVValueProperty(PROP_PVVALUE, null));
         addProperty(new IntegerProperty(PROP_ROW_COUNT, "Row Count",
-                WidgetPropertyCategory.Behavior, /*default=*/1, /*min=*/1, /*max=*/MAX_ROW_COUNT));
+                WidgetPropertyCategory.Behavior, /*default=*/1, /*min=*/1, /*max=*/200000));
```
willrogers, line 75: @rjwills28 200000 not MAX_ROW_COUNT?

rjwills28: Yes the 200000 is correct, it cannot be MAX_ROW_COUNT. This is a bit of a dummy value that we instantiate the widget with, allowing the max row count to be arbitrarily high but later ensuring that the MAX_ROW_COUNT is adhered to (i.e. delete the excess rows and log a message).

The problem is that if we use MAX_ROW_COUNT as the max when we add the property, then any row number above this gets lost, i.e. we don't even register it or keep account of it and so it gets painted at the top of the widget as we can't catch it to remove it. By adding the property with the 200000 we can keep account for all the rows that are above the MAX_ROW_COUNT (i.e. 200) and delete them so that they are not displayed. 

I can define this number as a constant somewhere if that would be clearer?

willrogers: Thanks, yes that would be clearer.

shroffk (approved):

PR #2699, opened 2021-07-08
Css564 annotation units fix

willrogers:
The previous change added the units at the end of the format string, not necessarily attached to the value.

This corrects that bug.

@rjwills28


willrogers:
cc @utzeln

kasemir (approved):

PR #2701, opened 2021-07-20
Fixing plotbins preference limits.

willrogers:
It isn't possible to enter the special value of '-3' into the databrowser preferences page.

This commit fixes that.

cc @rjwills28


kasemir (approved):
Is only -3 allowed? In the phoebus implementation, any negative value represents "abs(N) times the display width", so on a screen that's 1000 pixel wide, -3 would request 3000 bins. https://github.com/ControlSystemStudio/phoebus/blob/2f5de05be5f2827f095097a7963114b17a4efb31/app/databrowser/src/main/resources/databrowser_preferences.properties#L57

kasemir:
Meaning: Maybe "-1",  "-2" or "-4" should also be allowed? Unclear where to draw a limit. "-10" for 10 times the screen resolution would be mad, so limit it to say -5??

willrogers:
So it is. How did we not notice that?

willrogers:
Logic updated to reflect that misunderstanding.

kasemir:
I think that's good!

PR #2702, opened 2021-07-21
Css567 show trace bug

willrogers:
A previous change introduced a bug: traces may be hidden in a plot file but if you then try to show them they do not appear.

Becky fixed the problem, but I wondered if there was a simpler way of doing this: always add all traces to the plot and only show them if they are visible.

We need a review from @kasemir: it might be that you think this is memory hungry, for example.

cc @rjwills28


kasemir (approved):
Certainly looks easier to maintain when you keep the trace in the plot

PR #2703, opened 2021-07-30
A formula trace does not get immediately added to the databrowser and also does not immediately updated

rjwills28:
We came across an issue when users were adding formula traces to the data browser that they did not appear immediately in the plot. Equally if a user updates the formula, the changes do not get immediately displayed. The fomula only gets re-evaluated when the samples change, i.e. if the live PV variable changes or the received archived data changes.

I have created a fix for this issue which identifies when the formula has been updated and registers this change causing the formula to be re-evaluated.


willrogers:
@kasemir is this present in Phoebus?

kasemir (approved):

kasemir:
No, something like this is also needed in phoebus. https://github.com/ControlSystemStudio/phoebus/issues/1972

Issue #2717, opened 2022-08-24
cs-studio crashing on first start


codedump:
Hello,

I'm attempting to start cs-studio on a fresh [Fedora Silverblue](https://silverblue.fedoraproject.org/) [toolbox](https://containertoolbx.org/) (think of it as of a newly installed Fedora Linux system). Essentially, the steps I do after download and unpacking are these:
```
$ dnf install java-openjdk-latest
$ ./cs-studio
```

It crashes and leaves behind the following log file contents in `~/.cs-studio/.metadata/.log`:
```
SESSION 2022-08-24 15:31:59.299 -----------------------------------------------                                                                                                     
eclipse.buildId=unknown                                                                                                                                                              
java.version=18.0.2                                                                                                                                                                  
java.vendor=Red Hat, Inc.                                                                                                                                                            
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en                                                                                                                           
Command-line arguments:  -os linux -ws gtk -arch x86_64                                                                                                                              
                                                                                                                                                                                     
!ENTRY org.eclipse.equinox.ds 1 0 2022-08-24 15:32:00.000                                                                                                                            
!MESSAGE Could not bind a reference of component org.csstudio.diirt.util.RegisterDatasource. The reference is: Reference[name = DataSourceProvider, interface = org.diirt.datasource.
DataSourceProvider, policy = dynamic, cardinality = 0..n, target = null, bind = registerDatasource, unbind = deregisterDatasource]                                                   
                                                                                                                                                                                     
!ENTRY org.eclipse.equinox.ds 1 0 2022-08-24 15:32:00.001                                                                                                                            
!MESSAGE Could not bind a reference of component org.csstudio.diirt.util.RegisterFormulaFunctionSet. The reference is: Reference[name = FormulaFunctionSet, interface = org.diirt.dat
asource.formula.FormulaFunctionSet, policy = dynamic, cardinality = 0..n, target = null, bind = registerFormulaFunctionSet, unbind = deregisterFormulaFunctionSet]                   
                                                                                                                                                                                     
!ENTRY org.eclipse.equinox.ds 1 0 2022-08-24 15:32:00.001                                                                                                                            
!MESSAGE Could not bind a reference of component org.csstudio.diirt.util.RegisterService. The reference is: Reference[name = ServiceProvider, interface = org.diirt.service.ServicePr
[...]

PR #2718, opened 2022-09-20
#2717 adding a note about JDK version

shroffk:


shroffk:
@kasemir could you please review this

kasemir (approved):
I no longer have any experience with building the Eclipse sources, and the last binaries we ran here were 32 bit JDK 8...

PR #2720, opened 2023-05-12
Fix build on the latest version of Java 11

rjwills28:
I noticed that CS-Studio no longer builds on the latest version of Java 11 used by the CI job (Java 11.0.19). I investigated and as of Java 11.0.17 the version of JaCoCo that we used for tests and code coverage is no longer supported. 
I updated the version of JaCoCo to 0.8.3, which officially supports Java 11. This led to a few issues with some of the tests, which I have fixed so that CS-Studio now builds successfully and all of the tests pass.


rjwills28:
Hi @kasemir & @shroffk - would you mind having a look at this PR to fix the CI build on the latest Java version of 11? Changes only involve the running of the tests.

shroffk (approved):

PR #2725, opened 2024-02-27
Resize the legend area when traces are hidden

rjwills28:
We came across an issue in the databrowser where the area taken up to display the traces legend does not shrink when traces are hidden. This is particular noticeable if the plot contains many traces of which most are hidden. 
It turns out that in the code the legend area is determined using all the traces even if they are not being displayed. This PR fixes this issue by only considering the visible traces.


kasemir:
Turns out the JFX version had the 'continue' part but still counted all traces, so thanks for this update

shroffk:
@rjwills28 can you update this branch... with #2726 merged we should be able to get the CI working again here

PR #2729, opened 2024-05-13
Update JCA version to 2.4.10

rjwills28:
The latest release of JCA fixes an issue where the code could block during a socket connection by adding a configurable timeout. This would be a useful fix to have in CS-Studio. PR is to update the JCA dependency to the latest JCA version (2.4.10).


rjwills28:
Hi @shroffk , @kasemir - would it be possible to get a quick review on this PR to get it merged. It is just updating the JCA version. Many thanks!

shroffk (approved):

PR #2733, opened 2025-06-10
Identify thread pool alarms

Chsudeepta:
This is a follow up from the comment for [PR](https://github.com/ISISComputingGroup/cs-studio/pull/18) on naming the thread pool.


Chsudeepta:
Not required here

